PR #9, opened 2026-02-12
Fixed skolemization bug

amarshah1:
*Description of changes:* I fixed the bug relating to skolemization. We were calculating the wrong polarity in `quantifier.rs`. We now pass all regression tests


HuStmpHrrr:
closing it as it's in #7 already. could you help figure out a way to make z3 compile in CI?

Issue #13, opened 2026-02-14
Trouble building Z3-sys with older version of GCC


amarshah1:
I am getting this error when running cargo build with GCC 8.5.0. I switched to a newer version of GCC (13.3.1) and the error goes away. I am not sure what the best way to handle this is. Should we just add a specific gcc requirement?

Also, building z3 is quiet slow and would not be necessary all the time. There are two options to fix this

1. Allow the user to not build z3 and instead provide a path to their z3 header file
2. Allow the user to not build z3 at all if they just want to use our arithmetic solver. We need to make albion the default arithmetic solver before we do this.
3. Maybe there is a way to not have to build all of z3/build z3 in parallel

```
error: failed to run custom build command for `z3-sys v0.10.5`

Caused by:
  process didn't exit successfully: `/ocean/projects/cis230065p/ashah12/sundance/Sundance-SMT/target/debug/build/z3-sys-70d63f3df665c1d6/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  downloading to /ocean/projects/cis230065p/ashah12/sundance/Sundance-SMT/target/debug/build/z3-sys-194bee6e2b797614/out/z3
  Downloaded 7.14MB
  CMAKE_TOOLCHAIN_FILE_x86_64-unknown-linux-gnu = None
  CMAKE_TOOLCHAIN_FILE_x86_64_unknown_linux_gnu = None
  HOST_CMAKE_TOOLCHAIN_FILE = None
  CMAKE_TOOLCHAIN_FILE = None
  CMAKE_GENERATOR_x86_64-unknown-linux-gnu = None
  CMAKE_GENERATOR_x86_64_unknown_linux_gnu = None
  HOST_CMAKE_GENERATOR = None
  CMAKE_GENERATOR = None
  CMAKE_PREFIX_PATH_x86_64-unknown-linux-gnu = None
  CMAKE_PREFIX_PATH_x86_64_unknown_linux_gnu = None
  HOST_CMAKE_PREFIX_PATH = None
  CMAKE_PREFIX_PATH = None
  CMAKE_x86_64-unknown-linux-gnu = None
  CMAKE_x86_64_unknown_linux_gnu = None
  HOST_CMAKE = None
  CMAKE = None
  -- The CXX compiler identification is GNU 8.5.0
  -- Detecting CXX compiler ABI info
  -- Detecting CXX compiler ABI info - done
[...]

Issue #15, opened 2026-02-14
z3 build is unstable


HuStmpHrrr:
It seems that current z3 build is not stable and it seems that the build script is not using the z3 submodule in the repo.

HuStmpHrrr:
I think #17 stabilizes it.

PR #14, opened 2026-02-14
Fix the types assigned to arithmetic variables

benjaminfjones:
*Issue #, if available:*

*Description of changes:*

Incorrect types were being assigned to integer variables due to a combination of Yaspar updates (sort info moving) and a default fall-through type being in place. The fall-through has been removed and the types are now matched from the correct Yaspar AST field.

The arithmetic unit tests that were previously ignored are now enabled.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #16, opened 2026-02-14
Adding option to use local z3

amarshah1:
*Issue #, if available:* This partially resolves issues #13 and #15 by providing users the ability to use a local version of z3 instead of compiling from source.

*Description of changes:* Gives the user the ability to use a local version of z3 by including the flags `--no-default-features --features local-z3`. Updated README with new insstructions

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -92,7 +92,7 @@ jobs:
             ${{ runner.os }}-solver-
 
       - name: Run tests
-        run: cargo test --release --no-default-features --features "${{ matrix.features }}"
+        run: cargo test --release --no-default-features --features "${{ matrix.features }}" -- --nocapture
 
   lints:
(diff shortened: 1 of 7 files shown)


amarshah1:
This is not a complete fix for #13 and #15, ideally we would like z3 to compile from source on all machines. I will also make Sundance's arithmetic solver the default once #14 lands and so we can add an option to build with no z3

.github/workflows/build.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
         os: [ ubuntu-latest, macos-latest ]
         rust: [ stable ]
         features:
-          - ''
```
HuStmpHrrr: maybe set `default` instead?

amarshah1: sure, done

Issue #33, opened 2026-02-23
Bug with internal integer arithmetic solver


amarshah1:
When running the internal arithmetic solver on this query, I get `sat`. However, the correct answer is `unsat` which the z3 arithmetic solver gets.

```
(declare-sort P 0)
(declare-fun S (Int Int) Int)
(declare-fun v () Int)
(declare-fun b () Int)
(assert (< b 0))
(assert (= (- 1) (- v 1)))
(assert (= (- 1) (S 0 1)))
(assert (> b (S v 1)))
(check-sat)
```

PR #34, opened 2026-02-23
Fix sign in calculate_d0

leeabarnett:
*Issue #, if available:* https://github.com/yaspar-org/Sundance-SMT/issues/33

*Description of changes:* 

Thanks to the impeccably documented and cleanly structured code in lra_solver.rs, I was able to dig around (also with the help of AI) for some possible causes for the bug reported in the issue. High-level understanding of what's happening on this input: 

```
(declare-const b Int)
(assert (< b 0))
(assert (<= (- b) 1))
(assert (< (- b) 1))
```

1. LRA solver finds a result over $\mathbb{Q}_{\delta}$, which looks like $b = 0 - \delta$.
2. This needs to be converted to a solution for $b$ in $\mathbb{Q}$. This requires calling `calculate_d0` to find a value for $\delta$. It's coming up with $\delta = -1$ so that $b = 1$.
3. Now we have $b \in \mathbb{Z}$ so it looks like a valid solution.
4. We get the result `sat`. 

Fortunately `calculate_d0` has a really thorough docstring which I was reading over to try and understand how it works. This part looked like maybe an issue:

```
    /// On the other hand, if a bound is purely rational (i.e. non-strict in the original system),
    /// but the assigned value is not, we must ensure that instantiating δ_0 doesn't cause the
    /// assigned value to stray outside the inclusive bounds.
    ///
    ///   l_i <= x_i^rat + x_i^inf δ_0
    ///   0 <= (x_i^rat - l_i) + x_i^inf δ_0
    ///
    /// If x_i^rat = l_i but x_i^inf < 0 the Q_δ inequality would be violated, so WLOG x_i^rat -
    /// l_i > 0. If x_i^inf > 0 there is no constraint on δ_0 and finally if it's < 0 we can
    /// choose:
    ///
    ///   δ_0 <= (l_i - x_i^rat) / abs(x_i^inf)
```

If we're assuming `x_i^rat - l_i > 0` then wouldn't `l_i - x_i^rat < 0` ? which would mean the <= bound on delta_0 in the last line would be negative.

When I flip this sign in the code, the problem seems to be resolved, which would be great. However, I'm wary of this one-line change that doesn't affect any other tests. I think we need a better way to validate that this is correct, or at least I want to get some eyes on this.

Please let me know what you think!

---

### DIFF
--- a/src/arithmetic/lia/lira_solver.rs
+++ b/src/arithmetic/lia/lira_solver.rs
@@ -756,4 +756,29 @@ mod tests {
         let result = solve_smtlib(smt_conflict).expect("solver failed");
         assert!(matches!(result, SolverDecisionApi::FEASIBLE(_)),);
     }
+
+    // Regression test from https://github.com/yaspar-org/Sundance-SMT/issues/33
+    //
+    //  b <  0       {       b < 0
+    // -b <= 1  -->  { -1 <= b
+    // -b <  1       { -1 <  b
+    //
+    // The first and 3rd constraints are mutually UNSAT.
+    #[test]
+    fn regresssion_issue_33() {
+        let smt = r#"
+        (declare-const b Int)
+        (assert (< b 0))
+        (assert (<= (- b) 1))
+        (assert (< (- b) 1))
+        (check-sat)
+        "#;
+        let result = solve_smtlib(smt).expect("solver failed");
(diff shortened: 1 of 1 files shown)


benjaminfjones (approved):
The sign change looks good. Someone else should please review my additional comment changes and the unit test.

amarshah1:
Changes and comments look good to me!

PR #36, opened 2026-02-24
Add initial sparse matrix pivot algorithm

benjaminfjones:
*Description of changes:*

This implementation is intended to be straight forward (although lengthy) and not optimally efficient. It serves as a steel thread on which tests can be written and the API figured out.

There are well-known optimizations that will be added later.

### DIFF
--- a/src/arithmetic/lia/sparse.rs
+++ b/src/arithmetic/lia/sparse.rs
@@ -22,11 +22,12 @@
 
 use num_traits::Zero;
 use slotmap;
+use std::collections::{HashMap, HashSet};
 use std::error;
 use std::fmt;
 
 /// Generic error type for spare matrix operations
-#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


HuStmpHrrr:
you can convert the PR to a draft so it won't get merged.

src/arithmetic/lia/sparse.rs
```diff
@@ -22,11 +22,12 @@
 
 use num_traits::Zero;
 use slotmap;
+use std::collections::{HashMap, HashSet};
 use std::error;
 use std::fmt;
 
 /// Generic error type for spare matrix operations
-#[derive(Debug)]
```
HuStmpHrrr: is it intentional to avoid `Clone` and `Hash`?

HuStmpHrrr (approved):

Issue #39, opened 2026-03-02
Optimize datatype handling


HuStmpHrrr:
I think I see a few opportunities to simplify datatype handling. I will try to make some improvements.

PR #43, opened 2026-03-03
Add support for polymorphic datatypes

HuStmpHrrr:
related issue #39

*Issue #, if available:*

*Description of changes:*

This is a sequence of changes that improve datatype handling. This PR is intended to support polymorphic datatypes by taking sort parameters into account. 

### DIFF
--- a/src/egraphs/congruence_closure.rs
+++ b/src/egraphs/congruence_closure.rs
@@ -3,11 +3,13 @@
 
 //! Classic congruence closure algorithm
 
-use crate::datatypes::axioms::{learn_ctor_selector_clauses, learn_or_not_term_tester_term};
+use crate::cnf::CNFConversion as _;
+use crate::datatypes::axioms::learn_or_not_term_tester_term;
 use crate::egraphs::proofforest::ProofForestEdge;
+use crate::preprocess::check_for_function_bool;
 use crate::utils::{DeterministicHashMap, DeterministicHashSet, fmt_termlist};
-use yaspar_ir::ast::alg::CheckIdentifier;
-use yaspar_ir::ast::{FetchSort, IdentifierKind, Repr, Term, TermAllocator};
+use yaspar_ir::ast::alg::{CheckIdentifier, QualifiedIdentifier};
+use yaspar_ir::ast::{CheckedApi, FetchSort, IdentifierKind, Repr, Term, TermAllocator};
 
 use crate::egraphs::datastructures::{Assertion, ConstructorType::*, DisequalTerm, Predecessor};
 use crate::egraphs::egraph::Egraph;
@@ -115,7 +117,7 @@ pub fn process_assignment(
         } => {
             // need to add isC(n) => n = C(C^1(n),..., C^m(n))
             let dt_sort = inner_term.get_sort(egraph);
-            let _term_lit = egraph.get_lit_from_term(&term);
+            let term_lit = egraph.get_lit_from_term(&term);
             debug_println!(19, 0, "trying to get for the term {}", inner_term);
             match egraph.term_constructors.get(&inner_term.uid()).unwrap() {
                 Constructor {
@@ -140,6 +142,7 @@ pub fn process_assignment(
                             egraph,
                             tester_term.clone(),
                             term.clone(),
+                            true,
                         );
                         Some(tester_cnf)
                     }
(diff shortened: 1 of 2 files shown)


src/egraphs/congruence_closure.rs
```diff
@@ -160,56 +160,33 @@ pub fn process_assignment(
...
-                            let sel_id = QualifiedIdentifier::simple(sym.clone()); // todo: maybe just store this in egraph, so I don't have to recompute
-                            let sel_app = egraph.context.app(
-                                sel_id,
-                                vec![inner_term.clone()],
-                                Some(sort.clone()),
-                            ); // todo: not sure if I should give the sort here
+                        for sel in &ctor.args {
+                            let sel_app = egraph
+                                .context
+                                .typed_simp_app(sel.0.clone(), vec![inner_term.clone()])
```
HuStmpHrrr: this is the only place where checked API is used. it probbaly indicates a way to improve the current low level API by exposing some unification details.

amarshah1: you mean because we need to use a string as the function name?

src/preprocess.rs
```diff
@@ -290,82 +268,59 @@ fn find_datatype_axioms(
...
-                    .app(ctor_id, selectors_apps.clone(), Some(ctor_sort))
+                egraph.app(ctor_id, selectors_apps, Some(sort.clone()))
             };
 
-            for (i, sel_app) in selectors_apps.clone().iter().enumerate() {
-                // include new constraints for subterms
-                let sort = selector_sorts[i].to_string();
-                if egraph.datatype_info.sorts.contains_key(&sort) {
-                    let additional_constraints = find_datatype_axioms(sel_app, sort, egraph, false);
-                    vector.extend(additional_constraints.clone());
```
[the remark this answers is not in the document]
HuStmpHrrr: They are included a few lines above

src/egraphs/congruence_closure.rs
amarshah1: This is when you have parametricity? I am still a little confused. Whats an example that would require unification?

src/preprocess.rs
```diff
@@ -175,109 +161,101 @@ pub fn check_for_function_bool(
...
-            let tester_term = egraph
-                .context
-                .app(tester_qid, vec![term.clone()], Some(bool_sort));
+            // insert (_ is X)
+            let tester_term = egraph.app(
+                tester_identifier.into(),
+                vec![term.clone()],
+                Some(bool_sort.clone()),
+            );
+            // todo: we are not handling is-X here
```
benjaminfjones: I'm fine leaving this style tester out. I asked Andy about them and he said that it's a non-standard extension of SMT-LIB.

src/preprocess.rs
```diff
@@ -175,109 +161,101 @@ pub fn check_for_function_bool(
...
+    let is_symbol = egraph.allocate_symbol("is");
+
     // Step 1. Store the constructor in term_constructors
     let num = term.uid();
     if egraph.datatype_axioms_applied.contains(&num) {
         return vector;
     }
     egraph.datatype_axioms_applied.insert(num);
 
+    // todo: why do we need this? missing Global case?
```
HuStmpHrrr: @amarshah1 could you take a look at this TODO?

src/egraphs/congruence_closure.rs
HuStmpHrrr: This is true. This also requires yaspar to do the job. I will add a monomorphization function.

src/preprocess.rs
amarshah1: Yeah I don't think it is a bug because enums will always add the term: `is-C(C)`. But let me think about it more carefully in the morning

HuStmpHrrr:
> Added the refactoring. I want to also add helper functions for the datatype calls in `congruence_closure`. I will do that tomorrow morning. @HuStmpHrrr could you also add test cases with parametric datatypes

ok I will take it from here.

src/proof/proof_tracer.rs
```diff
@@ -132,36 +132,16 @@ fn format_datatype_declaration(sorts: &HashMap<Str, SortDef>) -> (String, HashSe
         match sort_def {
             SortDef::Opaque(..) | SortDef::Transparent { .. } => {}
             SortDef::Datatype(data) => {
-                sort_str.push(format!("({} 0)", sort_name));
+                sort_str.push(format!("({} {})", sort_name, data.params.len()));
```
HuStmpHrrr: ~~This again missed the polymorphic part of datatypes. will push a fix.~~ ah my bad

amarshah1:
I reverted some of the changes. There is still the bug when `lazy_dt` is off, but I can handle this in a separate PR

Issue #32, opened 2026-02-23
Bug related to how we backtrack when introducing new terms via datatypes


amarshah1:
Sundance returns `sat` on this problem, while the correct answer should be `unsat`. I believe this is related to a bug related to how new terms created by datatypes or quantifier instantiations can introduce new equalities that we have to backtrack on. I will push a fix for this soon.

```
(declare-sort P 0)
(declare-fun % (P) Int)
(declare-sort l 0)
(declare-datatypes ((c 0) (u 0)) (((co) (e (or P))) ((u))))
(declare-fun c (c) P)
(declare-fun t (u) P)
(declare-fun P (c) P)
(declare-fun o (P) c)
(declare-fun %P (P) u)
(declare-fun l (P) l)
(declare-const tm P)
(declare-const m c)
(assert (and (= m (o tm))))
(assert (or ((_ is co) m) (and (= (l (t (%P (c (o (P m)))))) (l (t (%P (c (o (c co))))))) (not (= (% (t (%P (c (o (c co)))))) (% (t (%P (c (o (c co)))))))))))
(assert ((_ is e) (o (P co))))
(assert (= co (o (P m))))
(check-sat)
```

PR #59, opened 2026-03-06
Fixing backtracking so that equalities from new terms are retained correctly after backtracking

amarshah1:
*Issue #, if available: #32 * 

*Description of changes: Changing how we do backtracking so that equalities from new terms learnt a level > 0 are retained after backtracking*


amarshah1:
The two main changes are 

1. Removing the `from_quantifier_backtrack_set`
2. Removing the break from the `union_eclass` function

PR #61, opened 2026-03-10
Fix unsat core production in lialp.rs

leeabarnett:
*Issue #, if available:*

*Description of changes:* Fixes a bug in how "Infeasible" results from the internal arithmetic solver are turned into unsat cores.

In `check_integer_constraints_satisfiable_lia` (in the internal arithmetic solver path, so in lialp.rs), we add an expression to the internal solver corresponding to each root in the egraph, and an expression for each constraint as produced by `extract_linear_constraints` in lp.rs. We are tracking the literals that are responsible for those constraints, and use that mapping to create the unsat core in case of an "infeasible" result from the arithmetic solver. 

We should also be tracking literals responsible for those roots, which are returned already by `extract_linear_expression` in lp.rs. Without them the unsat core can be too strong.

I found this by running some differential testing to compare using the internal arithmetic solver vs using z3. An example formula that triggered this is added as a new regression test.

---


amarshah1:
What do you mean by the "unsat core is too strong". Are you saying that prior to this fix, we were learning conflict clauses that were larger than necessary?

amarshah1:
Also, it might be worth it to spot check a few proofs with valido to make sure there are no introduced soundness issues. But otherwise it looks good!

leeabarnett:
> What do you mean by the "unsat core is too strong". Are you saying that prior to this fix, we were learning conflict clauses that were larger than necessary?

The conflict clauses are too small, meaning this is unsound. An example formula (where sundance with `--arithmetic internal` says unsat but the answer should be sat) is added as a test. In that formula, we have `(= 0 1)` and the literal for this is getting assigned to true. The egraph merges some terms based on this and we call the arithmetic solver. There we create an expression for each root in the egraph plus the constraints and correctly figure out that the system is infeasible. However nothing gets added to the unsat core to "blame" for the root expressions, since we're only tracking literals for the constraint slack variables. In this case we end up with an empty conflict clause at the end so we erroneously get unsat.

PR #63, opened 2026-03-10
Compare dense vs. sparse tableau pivot

benjaminfjones:
*Description of changes:*

This adds direct comparison tests between the dense tableau pivot operation and the sparse equivalent. Adding the tests surfaced the fact that the pivot row/col sign convention in the sparse implementation was flipped. This is fixed in the change.


amarshah1 (approved):
Looks good!

PR #68, opened 2026-03-13
Question: could we use IndexMap for egraph predecessors?

leeabarnett:
*Issue #, if available:*

*Description of changes:* I ran [samply](https://github.com/mstange/samply) on a few long running examples from the regression test suite. One thing I noticed was a decent amount of time spent in BTreeMap insert and remove, specifically on `predecessors` in the Egraph. This makes sense as updating this is surely part of the hot path for CC. However it made me wonder whether some other kind of map could be used. I assume we don't want to use HashMap so that we have deterministic ordering. From my understanding, BTreeMap order is based on the keys, while IndexMap order is based on insertion order. The keys in this case seem to be term IDs, so if these aren't assigned in any special way then maybe insertion order is fine here too?

I would be curious to know what others think. If there isn't an obvious reason this would kill stability, it seems to be worth benchmarking. It may not make a *huge* difference, but possibly noticeable.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ num-traits = "^0.2"
 serde_json = "1.0"
 slotmap = "^1.0"
 z3 = { version = "0.19.8", optional = true }
+indexmap = "2"
 num = "0.4.3"
 
 [dev-dependencies]
--- a/src/cadical_propagator.rs
+++ b/src/cadical_propagator.rs
@@ -15,8 +15,9 @@ use crate::log::is_important;
 use crate::proof::proof_tracer::SMTProofTracker;
 use crate::quantifiers::quantifier::QuantifierInstance::{Instantiation, Skolemization};
 use crate::quantifiers::quantifier::instantiate_quantifiers;
-use crate::utils::{DeterministicHashMap, DeterministicHashSet};
+use crate::utils::DeterministicHashSet;
 use cadical_sys::{CaDiCal, ExternalPropagator};
+use indexmap::IndexMap;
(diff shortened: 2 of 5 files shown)


HuStmpHrrr:
If we are really trying to tackle stability, then we should implement some normalization and use some variant of hash map for storing data. We can talk about it more definitely.

leeabarnett:
Agreed, reviewing data structures is good idea. I'll do a little more profiling to see what shows up.

> My understanding is IndexMap is "determinstic" based on the insertion order, so I was hopeful that a BTreeMap would be less sensitive to insertion order, i.e. if the assertions were swapped, a BTreeMap would be less sensitive to this.

This makes sense and matches my intuition as well. IndexMap preserves insertion order ([ref](https://github.com/indexmap-rs/indexmap)) while BTreeMap keeps keys ordered with respect to how they implement `Ord` ([ref](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html)).

In the case of `predecessors` the keys are `u64` term IDs, so a BTreeMap makes sure they are sorted with respect to the usual numeric ordering. From what I understand, replacing this with an IndexMap shouldn't make a huge difference in stability unless term IDs are assigned in some important way.

amarshah1:
But yeah I think we should merge this in as well. Do you think we should merge as is or if you want to switch out all DeterministicHashMap's and then merge? I think maybe the latter is better?

leeabarnett:
Closing this PR now that https://github.com/yaspar-org/Sundance-SMT/pull/83 was merged

PR #71, opened 2026-03-18
Adding basic statistic logging (right now only number of quantifier instantiations)

amarshah1:
*Issue #, if available:*

*Description of changes:*


benjaminfjones:
In the discussion today there were concerns about the signal handling part of the change. Can we keep the stats collection and printing and leave signal handling out for now? There was a suggestion to find a crate that handles this kind of functionality in a correct way.

benjaminfjones:
Here's a rust crate for posix signal handling: https://crates.io/crates/signal-hook It seems to address at least one concern that was brought up in our meeting today.

PR #74, opened 2026-03-20
Enable switching tableau implementations at runtime

benjaminfjones:
*Description of changes:*

This diff replaces the generic type parameter `T: Tableau` on `LRASolver<T>` and `LIRASolver<T>` with a runtime-dispatched enum `TableauImpl`. This lets the caller choose between dense and sparse
tableau representations at runtime via a `TableauKind` enum, rather than at compile time via generics. The original dense tableau is still the default implementation.

Three existing tests in `linear_system.rs` are duplicated to also run with `TableauKind::Sparse`, verifying the sparse backend produces the same results in those cases.

Replacing `TableauKind::Dense` with `TableauKind::Sparse` in the frontend, regression tests all pass and there is no regression in the number solved with `--arithmetic internal` set.

```
Test Summary:
Total tests: 358
Correct:     339
Incorrect:   0
Timeout:     19
test regression_test ... ok
```

### DIFF
--- a/src/arithmetic/lia/tableau.rs
+++ b/src/arithmetic/lia/tableau.rs
@@ -3,6 +3,8 @@
 
 //! Trait specifying common behavior for low-level tableaux
 
+use crate::arithmetic::lia::tableau_dense::TableauDense;
+use crate::arithmetic::lia::tableau_sparse::TableauSparse;
 use crate::arithmetic::lia::types::Rational;
 use std::error;
 use std::fmt;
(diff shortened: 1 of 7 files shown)


src/arithmetic/lia/tableau.rs
```diff
@@ -46,4 +65,88 @@ where
...
+    /// Dense tableau variant
+    Dense(TableauDense),
+    /// Sparse tableau variant
+    Sparse(TableauSparse),
+}
+
+impl PartialEq for TableauImpl {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (TableauImpl::Dense(a), TableauImpl::Dense(b)) => a == b,
```
HuStmpHrrr, line 91: is this equality intentional?

benjaminfjones: That's a good catch. I didn't implement the other cases. This equality is only used in tests. I don't intend to implement (or derive) equality on the sparse tableau directly but convert to dense first and then compare. I'll fix this.

benjaminfjones: done

HuStmpHrrr (approved):

PR #82, opened 2026-04-10
Add scripts and documentation for benchmarking

benjaminfjones:

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,54 @@
+# Scripts
+
+## run_verus.py
+
+Runs an SMT solver on every `.smt2` file in a directory (recursively) and
+categorizes each result as UNSAT, UNKNOWN, TIMEOUT, ERROR, or OTHER. Supports
+parallel execution and optional filtering against a previous results file to
+skip already-resolved benchmarks.
+
+### Usage
+
+```
+python scripts/run_verus.py <folder> --solver-command "<command>" [options]
+```
+
+### Options
+
+| Flag | Description |
+|---|---|
+| `--solver-command` | **(required)** Solver command to invoke on each file |
+| `-j`, `--jobs` | Number of parallel worker processes (default: 4) |
+| `--timeout` | Per-file timeout in seconds (default: 10) |
+| `-v`, `--verbose` | Print detailed stdout/stderr for each file |
+| `-o`, `--output` | Write categorized results to a file |
+| `--filter-from` | Path to a previous results file; only re-run files that were absent or had OTHER results |
+
+### Example
+
+```bash
+python scripts/run_verus.py verus_benchmarks/ \
+  --solver-command "/path/to/z3 auto_config=false" \
+  --timeout 10 -j 16 -o results.out 2>&1 | tee results.log
+```
+
+---
+
+## run_verus_all.sh
+
+Convenient wrapper that runs `run_verus.py` on four solver configurations in sequence:
+
+1. **Z3 4.15.4** with Verus-specific tuning flags
+2. **cvc5 1.3.3**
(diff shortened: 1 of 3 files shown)


scripts/run_verus.py
```diff
@@ -0,0 +1,400 @@
...
+    Args:
+        stdout: Standard output from cargo command
+        stderr: Standard error from cargo command
+    Returns:
+        Category: 'unsat', 'unknown', or 'other'
+    """
+    # Combine output for checking
+    combined_output = stdout.strip()
+    # Check for unsat result
+    if combined_output.endswith("unsat"):
```
HuStmpHrrr, line 80: there isn't sat?

scripts/run_verus_all.sh
```diff
@@ -0,0 +1,49 @@
...
+#   - Python with scripts/run_verus.py available
+#   - The benchmark directory (verus_benchmarks/) populated with .smt2 files
+#
+
+SMT_FOLDER="verus_benchmarks"
+# SMT_FOLDER="tests/regression/smt_files"
+NUM_WORKERS=16
+TIMEOUT=10
+
+Z3_COMMAND="/home/bfj/src/z3-z3-4.15.4/build/z3 auto_config=false smt.mbqi=false smt.case_split=3 smt.qi.eager_threshold=100.0 smt.delay_units=true smt.arith.solver=2 smt.arith.nl=false pi.enabled=false rewriter.sort_disjunctions=false"
```
HuStmpHrrr: do we plan to make it in the CI? In either case, should we invoke system binary, instead of pointing to your local binary?

benjaminfjones: I don't think so... running the whole benchmark set takes ~hours on a large machine. I can remove the path to my binaries and put a placeholder in if you think that's better.

scripts/run_verus.py
benjaminfjones: SAT is included with other in this categorization. That's because we expect UNSAT on all the benchmarks. We can break the categories down further and separate SAT out. I could do that in a subsequent PR.

scripts/run_verus_all.sh
HuStmpHrrr: yeah I think it's better to not point to a local path, or at least the script shouldn't only work in your env.

benjaminfjones: fixed

benjaminfjones:
CI is passing now

scripts/README.md
```diff
@@ -0,0 +1,54 @@
...
+| `-j`, `--jobs` | Number of parallel worker processes (default: 4) |
+| `--timeout` | Per-file timeout in seconds (default: 10) |
+| `-v`, `--verbose` | Print detailed stdout/stderr for each file |
+| `-o`, `--output` | Write categorized results to a file |
+| `--filter-from` | Path to a previous results file; only re-run files that were absent or had OTHER results |
+
+### Example
+
+```bash
+python scripts/run_verus.py verus_benchmarks/ \
```
HuStmpHrrr, line 30: I don't have this `verus_benchmarks/` folder. was it removed? any case, I don't have an issue and have approved.

HuStmpHrrr (approved):

scripts/README.md
benjaminfjones: Amar removed them before open sourcing. Check out the commit below and you should see the folder. If you want I can send you an archive instead.

https://github.com/yaspar-org/Sundance-SMT/commit/850a56fca0b2120198a85fd51755ebe0ac62b790

HuStmpHrrr: I am fine with it. maybe we should set up something internal?

PR #83, opened 2026-04-13
Implements better datastructures and faster implementation of union_predecessors

amarshah1:
*Issue #, if available:*

*Description of changes:* Switching predecessors to use a HashMap instead of BTreeMap. Also reducing the amount of cloning done in union_predecessors. This a major hot spot for the solver and these changes get us from solving 63.0% to 66.0% on Verus benchmarks. The datastructure changes are similar to those in #68 

### DIFF
--- a/src/egraphs/congruence_closure.rs
+++ b/src/egraphs/congruence_closure.rs
@@ -14,9 +14,7 @@ use yaspar_ir::ast::{
 };
 
 use crate::debug_println;
-use crate::egraphs::datastructures::{
-    Assertion, CanonicalOp, ConstructorType::*, DisequalTerm, Predecessor,
-};
+use crate::egraphs::datastructures::{Assertion, ConstructorType::*, DisequalTerm, Predecessor};
 use crate::egraphs::egraph::Egraph;
 use crate::egraphs::unionfind::ProofTracker;
 use crate::log::is_important;
@@ -1110,19 +1108,6 @@ fn union_predecessors(
     // you fix it, but then you compare to a for loop iterating through all terms in v and iteratively
     // computing the canonical_term_v, but this could change as you are iterating through the loop
     // so we want to precompute the canonical terms of v.
-    // Owned tuples here so we can restore predecessors_v before union_process_assignment
-    // runs (which can re-enter and read egraph.predecessors[v]).
-    // Precompute canonical forms for v's predecessors. Only store the key
-    // and the canonical form — predecessor data is looked up from the egraph
-    // later (after predecessors_v is restored), avoiding a Predecessor clone
-    // per entry.
-    let mut predecessor_v_canonical_forms: Vec<(u64, Option<(Vec<u64>, CanonicalOp, Vec<u64>)>)> =
-        Vec::with_capacity(predecessors_v.len());
-    for (pred_v_key, predecessor_v) in predecessors_v.iter() {
-        let canonical_form = egraph.get_canonical_form(predecessor_v.predecessor, level);
-        predecessor_v_canonical_forms.push((*pred_v_key, canonical_form));
-    }
-
     // moving predecessors from v to u
     for (pred_key, pred_val) in predecessors_v.iter() {
         debug_println!(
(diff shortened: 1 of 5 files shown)


src/egraphs/egraph.rs
```diff
@@ -1340,71 +1347,56 @@ impl Egraph {
...
-                self.get_term(new_pred_key),
-                new_pred_level,
-                new_pred_hash
-            );
+            Entry::Occupied(mut slot) => {
+                // Inline valid_hash for the original so we don't need &self inside
+                // the occupied borrow. Matches valid_hash's body exactly (minus
+                // its debug_println at level 5, which has no functional effect).
+                let original = slot.get();
+                let orig_valid = original.hash >= self.predecessor_level[original.level]
```
benjaminfjones: This matches valid_hash at line 1334 exactly, as you point out. But if valid_hash is ever updated, this copy will silently diverge. Consider extracting the logic into a free function or a #[inline] method on Predecessor that takes predecessor_level as a parameter, so there's a single source of truth.... especially in light of the TODO up there?

src/utils.rs
```diff
@@ -3,14 +3,24 @@
...
+// containers. Use this when you need canonical equality or sorted iteration.
 pub type DeterministicHashMap<K, V> = BTreeMap<K, V>;
 pub type DeterministicHashSet<T> = BTreeSet<T>;
 
+// Hash-based deterministic map using a fixed-seed FxHash. O(1) average ops
+// and deterministic across runs (unlike std's RandomState), but iteration
+// order depends on insertion sequence and these do not implement Hash.
+// Prefer this on hot paths where sorted iteration is not required.
+pub type FastDeterministicHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
+pub type FastDeterministicHashSet<T> = HashSet<T, BuildHasherDefault<FxHasher>>;
```
benjaminfjones, line 22: This doesn't appear to ever be used

src/egraphs/congruence_closure.rs
```diff
@@ -1022,29 +1026,35 @@ fn union_predecessors(
         )
     );
 
-    let predecessors_u = egraph.predecessors[u as usize].clone();
-    let predecessors_v = egraph.predecessors[v as usize].clone();
+    // Move u's and v's predecessor maps out of the egraph so we can iterate
+    // without cloning. Both slots are restored before any re-entrant call
+    // (add_predecessor / union_process_assignment) can observe them.
+    let mut predecessors_u = std::mem::take(&mut egraph.predecessors[u as usize]);
+    let predecessors_v = std::mem::take(&mut egraph.predecessors[v as usize]);
```
benjaminfjones, line 1038: Is there a bug here if `u == v`? It seems the second take will get a dummy value. And if so, what happens upon restore?

src/egraphs/congruence_closure.rs
```diff
@@ -457,7 +461,7 @@ fn leastcommonancestor_helper(
     let mut path_from_u = vec![];
     let mut curr = u;
 
-    if indent > 100 {
+    if indent > 1000 {
```
benjaminfjones: Can you add a justification for this magic number? Maybe also lift it to a named constant?

src/egraphs/congruence_closure.rs
benjaminfjones, line 1081: At some point we should remove commented out code like this. Not for this PR, just saying :)

src/egraphs/datastructures.rs
```diff
@@ -26,6 +26,21 @@ impl fmt::Display for DisequalTerm {
     }
 }
 
+/// Identifies the "operator" of a canonical term form for the purposes of
```
benjaminfjones, line 29: Nice! This is a great change

src/egraphs/congruence_closure.rs
amarshah1: Ok done! This is just a sanity check to make sure infinite loops dont happen

src/egraphs/congruence_closure.rs
amarshah1: u and v are guaranteed to not be equal. I will add an assert for this

src/egraphs/congruence_closure.rs
amarshah1: Yeah I can do a pass in a different PR dealing with commented out code.

src/utils.rs
amarshah1: Yeah its not, but I kept it because I thought it could be in the future? Ideally, I would like to switch all of the BTreeMap's and BTreeSet's to this faster version, but this requires a little bit more care

src/egraphs/egraph.rs
amarshah1: Ok, will do!

benjaminfjones (approved):

benjaminfjones:
I ran the regression tests under this PR and we now solve 2 more problems (on my machine), which is great! #timeouts went from 14 -> 12

```
Test Summary:
Total tests: 365
Correct:     353
Incorrect:   0
Timeout:     12
test regression_test ... ok
```

PR #86, opened 2026-04-16
Appease clippy for rust 1.95

benjaminfjones:


HuStmpHrrr:
a small optimization: clones of `original_subterms` are not necessary.

benjaminfjones:
> a small optimization: clones of `original_subterms` are not necessary.

It looks like there's a loop here where `original_subterms` is borrowed again later on subsequent iterations, so it can't be moved into the zip.

HuStmpHrrr (approved):

PR #89, opened 2026-04-25
Add SolverConfig and a default limit on branch-and-bound steps

benjaminfjones:
*Description of changes:*

This change adds arithmetic solver configuration. The configuration currently consists of the tableau kind (dense or sparse) and a new option to limit the number of branch-and-bound steps. Limiting branch-and-bound prevents it from diverging. When the limit is hit, the arithmetic solver returns UNKNOWN.

The default limit is chosen to be large enough so that all regression tests pass, i.e. either return the correct answer or timeout. Setting a much lower limit has the effect of turning some TIMEOUT  and UNSAT queries into UNKNOWNs.

Regression test results show one less timeout than before the change:

```
Test Summary:
Total tests: 365
Correct:     348
Incorrect:   0
Timeout:     17
test regression_test ... ok
```

On the Verus benchmarks, limiting branch-and-bound has the effect of raising the number of queries solved UNSAT from 56.3% to 58.8%.

----

### DIFF
--- a/src/arithmetic/lia/config.rs
+++ b/src/arithmetic/lia/config.rs
@@ -1,0 +1,28 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Solver configuration for the LIA arithmetic solver.
+
+use crate::arithmetic::lia::tableau::TableauKind;
+
+/// Configuration parameters for the LIA solver pipeline.
+#[derive(Debug, Clone)]
+pub struct SolverConfig {
+    /// Which tableau implementation to use (dense or sparse).
+    pub tableau_kind: TableauKind,
+    /// Upper bound on the number of LRA solve calls in branch-and-bound.
+    /// `None` means unlimited.
+    pub max_lra_solve_calls: Option<usize>,
+}
+
+impl Default for SolverConfig {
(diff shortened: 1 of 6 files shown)


src/arithmetic/lia/config.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+impl Default for SolverConfig {
+    fn default() -> Self {
+        Self {
+            tableau_kind: TableauKind::Dense,
+            // 2^17; high enough that regression tests pass as expected
+            // i.e. they are either SAT/UNSAT or TIMEOUT. If this is lowered
+            // significantly, some will return UNKNOWN instead.
+            max_lra_solve_calls: Some(131_072),
```
benjaminfjones, line 25: For now this is hardcoded. In the future it will be one of the heuristics you can configure on the command line.

amarshah1: Can you add this as an issue?

amarshah1 (approved):
Looks great, thanks!

src/arithmetic/lia/config.rs
benjaminfjones: Yes, will do!

benjaminfjones: Let me know if this captures it https://github.com/yaspar-org/Sundance-SMT/issues/91

PR #92, opened 2026-04-29
Fix non-basic pivot selection for Bland's rule

benjaminfjones:
*Description of changes:*

This is a small fix to the pivot selection rule. Previously the non-basic (column) variable selection followed column order rather than the fixed variable order. [Bland's rule](https://en.wikipedia.org/wiki/Bland's_rule) which we use to ensure termination of the simplex algorithm requires selection of the first eligible variable in the fixed order.

In practice, the bug does not affect any unit test or verus benchmark we target.

A regression test will be added in a separate PR.

### DIFF
--- a/src/arithmetic/lia/lra_solver.rs
+++ b/src/arithmetic/lia/lra_solver.rs
@@ -96,6 +96,8 @@ pub struct LRASolver {
     var_to_idx: BTreeMap<Var, usize>,
     /// Conversion context from the frontend including the name <-> Var mapping
     ctx: ConvContext,
+    /// During `solve()`, the current number of simplex steps that have been performed
+    num_simplex_steps: usize,
 }
 
 impl fmt::Debug for LRASolver {
@@ -111,6 +113,7 @@ impl fmt::Debug for LRASolver {
         s.push_str(&format!("Non-Basic pointers:\n  {0:?}\n", self.non_basic));
         s.push_str(&format!("Tableau:\n  {0:?}\n", self.tableau));
         s.push_str(&format!("State: {:?}\n", self.state));
+        s.push_str(&format!("Num simplex steps: {}\n", self.num_simplex_steps));
         // TODO: add the rest of the solver state, including backtracking info
         write!(f, "{}", s)
     }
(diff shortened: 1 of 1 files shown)


src/arithmetic/lia/lra_solver.rs
```diff
@@ -96,6 +96,8 @@ pub struct LRASolver {
     var_to_idx: BTreeMap<Var, usize>,
     /// Conversion context from the frontend including the name <-> Var mapping
     ctx: ConvContext,
+    /// During `solve()`, the current number of simplex steps that have been performed
+    num_simplex_steps: usize,
```
benjaminfjones, line 100: This is added for logging/stats purposes to confirm that Bland's rule selection does actually increase the number of simplex steps required in several examples.

src/arithmetic/lia/lra_solver.rs
```diff
@@ -530,36 +536,40 @@ impl LRASolver {
...
                 .upper
                 .as_ref()
                 .unwrap()
                 .clone(),
         };
 
-        for (col, non_basic_idx) in self.non_basic.iter().enumerate() {
-            let non_basic_var = &self.variables[*non_basic_idx];
+        // iterate over non_basic (col) variables in the fixed variable ordering
+        for var in self.variables.iter() {
```
benjaminfjones, line 558: The bug is fixed here

leeabarnett: Nice, so if I understand correctly the problem is that the order of `self.non_basic` is based on the columns and can change after pivots. We need follow a fixed order like in `self.variables` otherwise termination isn't guaranteed. Is that right?

leeabarnett (approved):
Looks good, how can we test it? Do you have a formula where this causes simplex to just spin?

src/arithmetic/lia/lra_solver.rs
benjaminfjones: Yep, exactly.

benjaminfjones:
> Looks good, how can we test it? Do you have a formula where this causes simplex to just spin?

I tried two examples from the literature that are cited as causing different pivot selection rules to not terminate, but in both cases the old code and the new code here converge. I suspect in the examples I tried that you have to follow both row and column order to get cycling, but in the old code here the basic variables are selected (correctly) according the fixed order.

I'm still looking for an example...

PR #94, opened 2026-04-30
Implement the SPASS-SATT heuristic for pivot variable selection

benjaminfjones:
*Description of changes:*

SPASS-SATT is an SMT solver that implements a dynamic pivot variable selection algorithm that accelerates the simplex procedure. See https://dx.doi.org/10.22028/D291-30636 section 7.1.1. It won the QF_LIA and QF_LRA divisions in SMT-Comp in 2018/2019.

Regression tests: one additional regression test is solved compared with `origin/main`. Results for the full verus benchmark set will be posted once they complete.

```
Test Summary:
Total tests: 367
Correct:     349
Incorrect:   0
Timeout:     16
test regression_test ... ok
```

----

### DIFF
--- a/src/arithmetic/lia/lra_solver.rs
+++ b/src/arithmetic/lia/lra_solver.rs
@@ -54,6 +54,15 @@ enum PivotDirection {
     Decrease,
 }
 
+/// Current pivot selection heuristic
+#[derive(Debug)]
+enum PivotHeuristic {
+    /// prefers basic variables by order and non-basic variables that are unbounded
+    Greedy,
+    /// prefers both basic and non-basic variables by order
+    Bland,
+}
+
 /// Linear real arithmetic solver
 pub struct LRASolver {
     /// Variable info for all original and slack variables. The vector itself should be immutable,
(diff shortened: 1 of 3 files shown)


src/arithmetic/lia/lra_solver.rs
```diff
@@ -54,6 +54,15 @@ enum PivotDirection {
     Decrease,
 }
 
+/// Current pivot selection heuristic
+#[derive(Debug)]
+enum PivotHeuristic {
+    /// prefers basic variables by order and non-basic variables that are unbounded
+    Greedy,
```
amarshah1, line 61: what is a (non)-basic variable?

src/arithmetic/lia/tableau.rs
```diff
@@ -137,6 +144,13 @@ impl Tableau for TableauImpl {
             TableauImpl::Sparse(t) => t.ncols(),
         }
     }
+
+    fn col_nnz(&self, col: usize) -> usize {
+        match self {
```
amarshah1, line 149: Am I missing something—why do you need this match statement?

amarshah1 (approved):
Looks good!

src/arithmetic/lia/lra_solver.rs
benjaminfjones: When the original system is setup, non-basic variables correspond to just ordinary variables in the problem. The basic variables correspond to slack variables. But after pivoting these get mixed up and so non-basic and basic get swapped around usually several times.

In terms of the tableau, non-basic variables correspond to columns and basic variables correspond to rows.

src/arithmetic/lia/tableau.rs
benjaminfjones: Good point. I think I can remove this entirely because there's a default impl in the trait. I'll do that in a subsequent PR.

PR #98, opened 2026-05-12
Implement the unit cube test

benjaminfjones:
*Description of changes:*

The unit cube test is described https://dx.doi.org/10.22028/D291-30636 section 4.3. It consists of finding a hypercube with unit side length inside the rational feasible set of a system of integer linear inequalities. If such a cube exists, then there is guaranteed to be an integer point in the feasible space, found by rounding the center of the hypercube to the nearest integer point.

This test is fast to perform and is implemented as a first step in the `LIRASolver::solve()` method. When it is successful, the arithmetic solver avoids branch and bound which can be expensive and/or diverge.

The regression tests that are added are the result of a bug found along the way.

Regression test results:

```
Test Summary:
Total tests: 367
Correct:     351
Incorrect:   0
Timeout:     16
test regression_test ... ok
```

----


src/arithmetic/lia/lira_solver.rs
```diff
@@ -532,6 +552,11 @@ impl LIRASolver {
     fn try_rounding_heuristic(&mut self) -> Option<Assignment<Var>> {
         self.lra_solver.try_rounding_heuristic()
     }
+
+    /// Attempt the unit cube test. Delegates to the underlying LRASolver.
```
amarshah1, line 556: What's the need for this function? Could we just call `self.lra_solver.try_unit_cube_test()` instead?

src/arithmetic/lia/lra_solver.rs
```diff
@@ -894,6 +912,133 @@ impl LRASolver {
...
+    /// Attempt the unit cube test
+    ///
+    /// The unit cube tests tightens basic variable bounds by (1/2)*|A_i|_1 and solves the rational
+    /// relaxation of the system. Here, |A_i|_1 means the sum of absolute values along the original
+    /// i-th tableau row. If the tightened system is feasible, rounding all integer variables to
+    /// the nearest integer is guaranteed to produce a valid integer solution.
+    pub fn try_unit_cube_test(&mut self) -> SolverResult<Option<Assignment<Var>>> {
+        let level = self.set_backtrack();
+
+        // Save tableau structure since solve() performs pivots that backtrack() does not undo
```
amarshah1, line 924: is there no way to avoid cloning here?

amarshah1: I guess you would need a way to backtrack on pivots, which is probably way beyond the scope of this PR.

amarshah1: But I think this would be necessary if we wanted a backtrackable arithmetic solver—maybe worth capturing this as an issue?

src/arithmetic/lia/lra_solver.rs
```diff
@@ -1657,4 +1802,213 @@ mod tests {
         // 7/2 is equidistant from 3 and 4; the implementation picks floor when tied
         assert!(*x_val == rbig!(3), "7/2 rounds to floor (3) on tie");
     }
+
```
amarshah1, line 1805: Tests look good to me!

amarshah1 (approved):

amarshah1:
I think this makes sense to me. I read the description [here](https://inria.hal.science/hal-01656397v1/document) about unit cubes and I think I understand it and your code

amarshah1:
I think before you merge it in, it might be good to get a better understanding of any performance improvements/regressions caused by this. I guess I am a bit worried because you could potentially have a regression if you spend time looking for a unit cube, but dont find one—do you ever experience anything like this?

src/arithmetic/lia/lra_solver.rs
benjaminfjones: Yeah. It's worth noting that cloning the tableau is about as expensive as a single pivot (for the dense tableau) so doing a single extra one before simplex starts adds negligible overhead.

I'll file and issue for backtrackable arithmetic.

src/arithmetic/lia/lira_solver.rs
benjaminfjones: Yes, good catch. I had more logic in here at one point and then removed it. I'll remove this in rev2

benjaminfjones:
> I think before you merge it in, it might be good to get a better understanding of any performance improvements/regressions caused by this. I guess I am a bit worried because you could potentially have a regression if you spend time looking for a unit cube, but dont find one—do you ever experience anything like this?

There is definitely overhead in trying to find a cube, however it amounts to a single simplex call which is very cheap compared to (potentially) exponentially many calls in branch-and-bound.

I've found that it's not increasing timeouts on our Verus benchmarks, nor on the regression tests. It improves the number of problems proved unsat by about 1.5% in the large Verus benchmark set. I've also run the code on arithmetic specific benchmarks that I produced using your logging code and I don't see any performance regressions there either.

Ultimately what I'm hoping is that combined with equality elimination, the unit cube test will help avoid branch and bound for SAT instances.

src/arithmetic/lia/lira_solver.rs
benjaminfjones: Actually, if it's okay with you, I'll remove the unneeded wrapper method and make some cleanups in a separate PR.

src/arithmetic/lia/lra_solver.rs
benjaminfjones: https://github.com/yaspar-org/Sundance-SMT/issues/103

PR #99, opened 2026-05-15
Implement equality elimination

benjaminfjones:
*Description of changes:*

Implement equality substitution as a pre-processing pass in the arithmetic component. The equalities eligible for substitution are of the form `x = y`, `x = C`, and `x = y + C` where `x`,`y` are variables and `C` is a rational constant.

The generation of conflicts and models is handled properly by tracking provenance of substitutions into relations.

Timeouts during regression tests have improved from 16 to 12:

```
Test Summary:
Total tests: 372
Correct:     360
Incorrect:   0
Timeout:     12
test regression_test ... ok
```

On the new LIA micro benchmarks, equality elimination shows a huge speedup compared to the current `main` branch; as much as 93% in one case:

```bash
$ cargo bench --no-default-features --bench lia_benchmarks
...
lia_00000_a1a014bc      time:   [1.2959 ms 1.2962 ms 1.2966 ms]
                        change: [+0.9713% +1.0085% +1.0484%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe

lia_00508_532fc04c      time:   [5.9540 ms 5.9554 ms 5.9570 ms]
                        change: [-86.594% -86.590% -86.587%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

lia_00301_06f02f64_unsat
                        time:   [16.423 ms 16.432 ms 16.441 ms]
                        change: [-93.579% -93.576% -93.572%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

lia_00301_6af0beef_unsat.smt2
                        time:   [24.649 ms 24.655 ms 24.662 ms]
                        change: [-87.967% -87.963% -87.960%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

----


amarshah1:
Can you run this on the Verus benchmark set to see if we see an improvement there as well?

amarshah1:
Does it every make sense to do more general constant folding? For instance `x = y + z`

tests/regression_test.rs
```diff
@@ -102,7 +102,10 @@ fn regression_test() {
             match wait_with_timeout(child, Duration::from_secs(10)) {
                 Ok(output) => {
                     let actual = String::from_utf8_lossy(&output.stdout).trim().to_string();
-                    if actual == expected {
+                    // Special case of unsat vs. unknown is to accommodate arithmetic benchmarks
+                    // where z3-arithmetic and internal arithmetic characteristics cause the solver
+                    // to differ in completeness.
+                    if actual == expected || (expected == "unsat" && actual == "unknown") {
```
amarshah1: what is an example of this?

src/arithmetic/lia/frontend.rs
```diff
@@ -551,24 +579,73 @@ pub fn solve_ctx_raw(ctx: &mut ConvContext, config: &SolverConfig) -> FrontendRe
...
         "lia::frontend::solve_ctx_raw: LIRA solver finished, duration: {duration:?}"
     );
-    ret
+
+    // Post-process solver result with equality elimination information
+    match ret.decision {
+        SolverDecision::FEASIBLE(ref mut model) => {
+            back_substitute_model(model, ctx.get_substitutions());
+        }
+        SolverDecision::INFEASIBLE(ref mut conflict) => {
```
amarshah1, line 635: To check if I am understanding this correctly, here we are taking the conflict that was produced and trying to resubstitute in equalities that were eliminated?

amarshah1: For instance if you had

x = 5
y < z + x
z < y - 4

If you substituted in x = 5 and then the LIA came up with a conflict clause "y < z + x" and "z < y - 4", how do we reincoporate x = 5 into the conflict?

benjaminfjones:
> Does it every make sense to do more general constant folding? For instance `x = y + z`

Yeah, I think so. I'd like to evaluate a few other types of substitution for how often they appear in our benchmarks and then generalize in a separate PR.

benjaminfjones:
> Can you run this on the Verus benchmark set to see if we see an improvement there as well?

Yeah, there is unfortunately a ~1% regression in the number of UNSAT instances we solve. The instances go from unsat to timeout in these cases just like the two regression tests I marked. I suspect this is related to #110 

Here are the results for Sundance w/ internal arithmetic on the Verus benchmarks on this branch:

```
Total files processed: 3763
  - UNSAT results:   2123 (56.4%)
  - UNKNOWN results:   10 (0.3%)
  - TIMEOUT results: 1586 (42.1%)
  - OTHER results:     44 (1.2%)

  - Neither UNSAT nor UNKNOWN: 1630 (43.3%)
```

and here is the baseline on the `main` branch

```
Total files processed: 3763
  - UNSAT results:   2177 (57.9%)
  - UNKNOWN results:   12 (0.3%)
  - TIMEOUT results: 1530 (40.7%)
  - OTHER results:     44 (1.2%)

  - Neither UNSAT nor UNKNOWN: 1574 (41.8%)
```

amarshah1:
I think #110 can cause unsat -> timeouts, but it should not cause an unsat -> unknown

benjaminfjones:
> I think #110 can cause unsat -> timeouts, but it should not cause an unsat -> unknown

Sorry, that's what I meant. I edited my last comment above but it probably doesn't come through by email. All the differences went from unsat -> timeout here.

src/arithmetic/lia/context.rs
```diff
@@ -65,6 +66,11 @@ pub struct ConvContext {
     relation_vars: Vec<Var>,
     /// All variables in the context, no matter how they are allocated
     all_variables: HashSet<Var>,
+    /// Substitutions applied during equality elimination, for model back-substitution
+    eliminated_substitutions: Vec<Substitution>,
+    /// Provenance: maps each relation's slack var to the set of equality slack vars
+    /// that have been substituted into it during equality elimination
+    provenance: HashMap<Var, BTreeSet<Var>>,
```
amarshah1, line 73: why do you use a BTreeSet here? Is it because you have to iterate over the set?

src/arithmetic/lia/equality_elim.rs
```diff
@@ -0,0 +1,690 @@
...
+                to_remove.insert(var_copy);
+            } else if rel.is_trivial_unsat() {
+                found_unsat = Some(var_copy);
+                break;
+            }
+        }
+
+        // Update provenance for all relations (the source equality's provenance
+        // is inherited by every relation it was substituted into)
+        for (_rel, var) in ctx.get_relations() {
```
amarshah1, line 295: this loop is confusing me a bit. Are you adding src_provenance to every possible relation? If so, isnt this doing a lot of redundant work?

amarshah1 (approved):
Approving so this isn't a blocker, but can you reply to my comments before merging in?

tests/regression_test.rs
benjaminfjones: Thanks for pointing this out; it's not needed anymore. An example is the subtraction regression tests that we expect unsat for (but always timeout). With this change, and on some platforms, unknown would be returned instead because Sundance is progressing faster and hitting the branch and bound limit.

I'll remove this.

src/arithmetic/lia/frontend.rs
benjaminfjones: That's what the provenance tracking does. When the subst occurs we track that x=5 was subsituted into the first relation and when that relation shows up in a conflict we add x=5 to the set.

src/arithmetic/lia/context.rs
benjaminfjones: I think because it's used to build the final conflict returned which is an ordered list of arithmetic relation terms. I will double check that.

src/arithmetic/lia/equality_elim.rs
benjaminfjones: Not to every possible relation, only those where `var` is `src_var`. I think the comment may be slightly misleading here.

Or are you asking why the src provenance set is being added to each relevant var's set when they could be shared? I think they could be shared with a better data structure here. In practice, I measured how large the provenance sets become in our benchmarks and even in large arithmetic problems (~100's of variables) they don't get larger than 10-20 ish? This is how deeply nested the chains of substitutions get.

tests/regression_test.rs
benjaminfjones: Done

PR #101, opened 2026-05-18
Add some basic stats

leeabarnett:
*Issue #, if available:*

*Description of changes:*

I was talking with Long and we thought it would be useful to have some way of tracking stats across a run. This is a start at that, with some very basic stats added. The main thing to figure out now is about whether this approach to tracking & reporting stats is what we want, since we can extend this to include more information in future PRs. 

Example:
```
 % ./target/release/sundance-smt tests/hardertests/unknown/unknown_reduced_znew.smt2 --stats
unsat
{
  "decisions": 4604,
  "backtracks": 345,
  "arith_checks": 8,
  "instantiations": 4914,
  "solve_time": 1.279
}
```

---

### DIFF
--- a/src/cdcl.rs
+++ b/src/cdcl.rs
@@ -7,6 +7,7 @@ use crate::cadical_propagator::CustomExternalPropagator;
 use crate::debug_println;
 use crate::egraphs::egraph::Egraph;
 use crate::proof::proof_tracer::SMTProofTracker;
+use crate::stats::SolverStats;
 use crate::utils::DeterministicHashSet;
 use cadical_sys::{CaDiCal, Status};
 use std::cell::RefCell;
@@ -28,7 +29,7 @@ pub fn cdcl_decision_procedure(
     symbol_table: HashMap<Str, Vec<(Sig, FunctionMeta)>>,
     arithmetic: ArithSolver,
     _timeout: u64, // todo: add timeout functionality
-) -> Status {
+) -> (Status, SolverStats) {
     let mut solver = CaDiCal::new();
 
     // Create proof tracker for real-time proof tracking wrapped in Rc<RefCell<>>
@@ -47,6 +48,7 @@ pub fn cdcl_decision_procedure(
         assignments: vec![0, 0],
         solver: &mut solver as *mut CaDiCal,
         arithmetic,
+        stats: SolverStats::new(),
     };
 
     solver.connect_external_propagator(&mut propagator);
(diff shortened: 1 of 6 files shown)


src/cdcl.rs
```diff
@@ -28,7 +29,7 @@ pub fn cdcl_decision_procedure(
     symbol_table: HashMap<Str, Vec<(Sig, FunctionMeta)>>,
     arithmetic: ArithSolver,
     _timeout: u64, // todo: add timeout functionality
-) -> Status {
+) -> (Status, SolverStats) {
```
benjaminfjones, line 32: Would be nice here to have a dedicated return type rather than a tuple. But that's something we can do later if wanted.

src/stats.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// Statistics about the solver run
+#[derive(Debug)]
+pub struct SolverStats {
+    start_time: Instant,
+
+    pub decisions: u64,
+    pub backtracks: u64,
+    pub arith_checks: u64,
+    pub instantiations: u64,
```
benjaminfjones, line 19: Docs on the fields would be nice

benjaminfjones:
I like the overall approach. It looks flexible. There are more granular arithmetic solver stats I'd like to fold in later and those could live in sub objects of the stats object if we like.

benjaminfjones (approved):

amarshah1:
This looks good to me. I think it might be useful to also add fields for 

1. SAT solver conflicts
2. Congruence closure conflicts

amarshah1:
As a further step (beyond the scope of this PR). I think it might be useful to have logging with all of this information, basically shows when the solver makes a decision, backtracks, adds quantifier instantiation, or finds a conflict, but chronologically. If we could get similar logging out of z3 it would be very helpful to compare the difference between the two

src/stats.rs
leeabarnett: Yes, good call, let me do that .

src/cdcl.rs
leeabarnett: Agreed that's a good idea for a follow-up. Thanks!

leeabarnett:
Thanks all!

> There are more granular arithmetic solver stats I'd like to fold in later and those could live in sub objects of the stats object if we like.

Great idea, we should expose the stats already collected by the internal arithmetic solver here in a future PR

> I think it might be useful to also add fields for SAT solver conflicts, Congruence closure conflicts

> As a further step (beyond the scope of this PR). I think it might be useful to have logging with all of this information, basically shows when the solver makes a decision, backtracks, adds quantifier instantiation, or finds a conflict, but chronologically. If we could get similar logging out of z3 it would be very helpful to compare the difference between the two

Agreed on all points! Can address in a follow-up.

src/stats.rs
leeabarnett: Done

PR #104, opened 2026-05-20
Avoid a clone and multiply in QDelta::instantiate

benjaminfjones:
*Description of changes:*

This is a minor performance improvement that avoids a clone and multiply on a hot path in the arithmetic solver.

----


amarshah1 (approved):
Looks great!

PR #106, opened 2026-05-22
gitignore chore and filter out tautologies in eDRAT proof.

amazoniancodel:
This PR contains two commits. One commit updates the .gitignore file to ignore eDRAT proof objects. The other adds a filtering function `clause_is_tautology()` to `proof_tracer.rs` to filter out added and deleted clauses that are tautologies. Deletions are also re-formatted to be "DRAT-style" where the entire clause is included, rather than referencing a clause ID, which is "LRAT-style."

Regarding the tautologies: For technical reasons due to Sundance's dependence on an outdated version of CaDiCaL/cadical-sys (see [this GitHub issue](https://github.com/arminbiere/cadical/issues/140)), tautologies are given to CaDiCaL during preprocessing of the SMT formula to force a complete model. (See [line 87 of preprocess.rs](https://github.com/yaspar-org/Sundance-SMT/blob/28bf8be05a2b94a767d7846150edb723cfe143a9/src/preprocess.rs#L87).) In my discussion with Amar Shah, we concluded that the easier thing was to filter out tautologies when storing eDRAT proof steps, but in the future, an upgrade to the underlying CaDiCaL version could remove the need to add tautologies, and thus eliminate the need for this kind of filtering.

------------

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,8 @@ target/
 .DS_Store
 .z3-trace
 .vscode
+
+# eDRAT proof files
+*.edrat
+*.drat
+*.lrat
--- a/src/cadical_propagator.rs
+++ b/src/cadical_propagator.rs
@@ -146,7 +146,7 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
                 negated_model_or_datatype_constraints_opt
             {
                 for constraint in negated_model_or_datatype_constraints {
-                    // todo: deleting this ordering thing -> just for debuggin
+                    // todo: deleting this ordering thing -> just for debugging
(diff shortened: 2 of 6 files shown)


amarshah1:
Did you figure out if newer versions of Cadical fixes this? If it does, it might just be worth it to move to the newer version of Cadical now, instead of adding more work-arounds

benjaminfjones (approved):
Looks good to me. I'm not an expert on the proof tracing components though, so it would be ideal if Amar can have a look as well.

amarshah1:
Looks good. We should try to figure out if updating our Cadical version can prevent us from relying on cadical-sys, but I don't think this should hold up this PR. I captures this here: #123

PR #111, opened 2026-05-26
Implement timeout

leeabarnett:
*Issue #, if available:*

*Description of changes:* Implementing `--timeout`. This is using a Cadical Terminator.

The reason I wanted to do this is so that you can now get output from `--stats` without waiting for a real result.

---


amarshah1 (approved):
Looks great!

benjaminfjones (commented):
Note that if the arithmetic solver gets stuck or takes a long time, the timeout won't be honored.

leeabarnett:
Yeah good point. And there are other places where this won't be checked, so it may not terminate in the expected time. It's a very soft time limit. We will want to improve this.

PR #114, opened 2026-05-28
Rewrite sparse matrix

benjaminfjones:
*Description of changes:*

This is a rewrite of the spare matrix module which underlies the sparse tableau and simplex algorithm. The goal is to simplify the code, in particular moving away from a pointer chasing implementation on top of an arena.

 The sparse matrix has been rewritten from a `slotmap`/circular-linked-list design to mirrored `Vec<HashMap<usize, V>>` row and column vectors. The pivot operation is now obviously terminating. Several small optimizations to arithmetic in the pivot have also been made.

  Changes:
  - src/arithmetic/lia/sparse.rs — complete rewrite (~220 lines vs ~987 lines before)
  - src/arithmetic/lia/tableau_sparse.rs — added efficient `col_nnz` override

  Results:
  - All 110 tests pass
  - Benchmark: sparse pivot is ~794.07 µs for 100 pivots (improved by ~40% !)
  - Code is dramatically simpler — no arena allocation, no pointer management, no circular linked lists

See the comments for regression test and Verus benchmark results.

----

### DIFF
--- a/src/arithmetic/lia/config.rs
+++ b/src/arithmetic/lia/config.rs
@@ -18,7 +18,11 @@ pub struct SolverConfig {
 impl Default for SolverConfig {
     fn default() -> Self {
         Self {
-            tableau_kind: TableauKind::Dense,
+            tableau_kind: TableauKind::Sparse,
+            // Note: 2^17 is high enough that regression tests pass as expected
+            // i.e. they are either SAT/UNSAT or TIMEOUT. If this is lowered
+            // significantly, some will return UNKNOWN instead since branch-and-bound
+            // will return UNKNOWN when the limit is hit.
             max_lra_solve_calls: None,
         }
     }
--- a/src/arithmetic/lia/sparse.rs
+++ b/src/arithmetic/lia/sparse.rs
@@ -1,14 +1,14 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Sparse matrix implementation using mirrored row/column HashMaps.
+//! Sparse matrix implementation using mirrored row/column FxHashMaps.
 //!
 //! Each entry (i, j, v) is stored in both `rows[i]` and `cols[j]`, enabling
 //! O(1) access by either row or column. This is the pattern used by most SMT
 //! simplex implementations (Yices2, Z3, CVC5).
 
 use num_traits::Zero;
-use std::collections::HashMap;
+use rustc_hash::FxHashMap;
 use std::error;
 use std::fmt;
 
@@ -31,8 +31,8 @@ pub type SparseResult<T> = Result<T, SparseError>;
 /// enabling O(1) access by either row or column during pivot operations.
 #[derive(Clone)]
 pub struct Matrix<V> {
-    rows: Vec<HashMap<usize, V>>, // rows[i]: col -> coefficient
-    cols: Vec<HashMap<usize, V>>, // cols[j]: row -> coefficient
+    rows: Vec<FxHashMap<usize, V>>, // rows[i]: col -> coefficient
+    cols: Vec<FxHashMap<usize, V>>, // cols[j]: row -> coefficient
     zero: V,
 }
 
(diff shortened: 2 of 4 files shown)


benjaminfjones:
Pivot benchmark compared to `main` branch

```
sparse_pivot_0_0_100x   time:   [793.89 µs 794.07 µs 794.23 µs]
                        change: [-39.157% -39.128% -39.098%] (p = 0.00 < 0.05)
                        Performance has improved.
```

benjaminfjones:
Regressions

```
Test Summary:
Total tests: 367
Correct:     351
Incorrect:   0
Timeout:     16
test regression_test ... ok
```

benjaminfjones:
Verus Benchmarks

```
Total files processed: 3763
  - UNSAT results:   2193 (58.3%)
  - UNKNOWN results:   12 (0.3%)
  - TIMEOUT results: 1514 (40.2%)
  - OTHER results:     44 (1.2%)

  - Neither UNSAT nor UNKNOWN: 1558 (41.4%)
```

% UNSAT is up from 57.9% on `main` branch

src/arithmetic/lia/sparse.rs
```diff
@@ -36,556 +22,224 @@ impl fmt::Display for SparseError {
...
+///
+/// Both `rows[i]` and `cols[j]` store the same coefficient for entry (i, j),
+/// enabling O(1) access by either row or column during pivot operations.
 #[derive(Clone)]
 pub struct Matrix<V> {
-    arena: slotmap::SlotMap<K, Node<V>>,
-    baserows: Vec<K>, // length = # rows
-    basecols: Vec<K>, // length = # cols
+    rows: Vec<FxHashMap<usize, V>>, // rows[i]: col -> coefficient
+    cols: Vec<FxHashMap<usize, V>>, // cols[j]: row -> coefficient
```
amarshah1, line 35: Is there a reason to not just do HashMap<usize, Hashmap<usize, V>>. Is the expectation that there are at least enough entries that you should have O(1) entry in each row and column?

src/arithmetic/lia/config.rs
```diff
@@ -18,7 +18,11 @@ pub struct SolverConfig {
 impl Default for SolverConfig {
     fn default() -> Self {
         Self {
-            tableau_kind: TableauKind::Dense,
+            tableau_kind: TableauKind::Sparse,
+            // Note: 2^17 is high enough that regression tests pass as expected
```
amarshah1, line 22: should you change this comment since there are no max_lra_solve_calls now

amarshah1:
Is slotmap ever used now? If its not, can we remove the import

src/arithmetic/lia/sparse.rs
```diff
@@ -36,556 +22,224 @@ impl fmt::Display for SparseError {
...
+    /// This is an O(1) operation.
     pub fn update_or_insert(&mut self, row: usize, col: usize, val: V) -> SparseResult<bool> {
-        if row >= self.baserows.len() || col >= self.basecols.len() {
-            return Err(SparseError(format!(
-                "({row}, {col}) are out of bounds for matrix of size {}x{}",
-                self.baserows.len(),
-                self.basecols.len(),
-            )));
-        }
+        self.validate_row_col(row, col)?;
```
amarshah1, line 115: The error message from validate_row_col says something about a pivot position, but I think this is not for a pivot position?

amarshah1 (approved):
Looks great overall! All my comments are pretty minor if you want to take a look at them and them merge?

src/arithmetic/lia/sparse.rs
benjaminfjones: The index of each outer vector is a row/col number of the full tableau, rather than row/col's of entries. In other words, each entry of the vector represents a whole row/col. So we could use hashmap at the outside, but the keys would always contain {0, ..., n-1}.

src/arithmetic/lia/config.rs
benjaminfjones: I left that in for the future in case we change back to `Some(n)`

src/arithmetic/lia/sparse.rs
benjaminfjones: Good catch, I'll fix that

benjaminfjones:
> Is slotmap ever used now? If its not, can we remove the import

It's still used in branch-and-bound (which I would also like to rewrite for simplicity in the future!)

PR #118, opened 2026-06-09
Generic trait for egraph

amarshah1:
*Issue #, if available:*

*Description of changes:* I added a generic trait for an egraph (`src/egraphs/trait.rs`) and refactored our current implementation to support this trait. This required me to split the old egraph into two, the `SolverState` (in `src/solver_state.rs`), which maintains high-level things the solver needs to keep track of, and the new `Egraph` (in `src/egraphs/basic/egraph.rs`), which does the egraph specific operations. The Egraph can only be called via the trait functions and maintains its own representation of terms as suggested by Bruno.

Around 80% of the code was writen using Claude Opus 4.6 while also manually approving every change myself and providing constant guidance. I was overall pretty happy with this approach. Claude sometimes got confused and made mistakes, so I think my guidance was necessary. I think this was faster and better than if I tried to do the refactoring completely on my own


amarshah1:
We regress from 66.1% to 65.5% on the Verus benchmarks. I anticipate this is largely because we waste time converting from the Yaspar representation to the egraph representation

src/egraphs/basic/egraph.rs
```diff
@@ -0,0 +1,2096 @@
...
+        }
+    }
+
+    /// Assert t1 = t2 at the given decision level.
+    /// Performs congruence closure. Returns a conflict if a disequality is violated.
+    fn assert_equal(&mut self, t1: u32, t2: u32, level: usize) -> EgraphResult<u32> {
+        self.advance_to_level(level);
+        let fixed = level == 0;
+        let from_quantifier = false;
+        let proof_parent = ProofForestEdge::Equality {
```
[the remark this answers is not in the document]
amarshah1: removed in new commit

src/egraphs/basic/egraph.rs
```diff
@@ -0,0 +1,2070 @@
...
+    type Op = Op;
+    type TermId = u32;
+
+    fn register_true(&mut self) -> Self::TermId {
+        self.register_opaque_term()
+    }
+
+    fn register_false(&mut self) -> Self::TermId {
+        self.register_opaque_term()
+    }
```
benjaminfjones: The egraph has no notion of which opaque ID is true vs false; that knowledge lives only in SolverState (true_uid/false_uid + bimap). This works today, but the trait's doc comment ("Register the boolean true constant") implies the egraph tracks them.

This all seems a bit fragile. I don't have a good suggestion to improve it, but documentation at this impl level would help clarify that polarity tracking is the caller's responsibility.

src/egraphs/traits.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub diseq_lit: Lit,
+}
+
+/// Result of a mutating egraph operation (assert_equal, assert_disequal, etc.).
+#[derive(Debug, Clone)]
+pub struct EgraphResult<T> {
+    /// A conflict, if the operation caused a disequality violation.
+    pub conflict: Option<Conflict<T>>,
+    /// SAT literals to propagate (from watched equalities that became true).
+    pub propagations: Vec<Lit>,
```
benjaminfjones, line 37: Is this "dead". It doesn't seem to be populated anywhere that I can see.

src/arithmetic/lp.rs
```diff
@@ -341,8 +342,11 @@ pub fn extract_linear_expression(
             }
         }
         Global(..) => {
-            // Handle variables
-            expr.insert(Coefficient::Term(term.uid()), IBig::from(1));
+            // TODO: consider whether we need egraph.find() here for correctness when variables are merged
```
benjaminfjones, line 345: Is this a good opportunity to just use `find`? Is there a good reason not to for consistency with the other match cases?

src/solver_state.rs
```diff
@@ -0,0 +1,938 @@
...
+                    triggers: trigger_ids,
+                    variables,
+                    body: inner_term.uid(),
+                    id: term.uid(),
+                    guard,
+                    polarity,
+                    skolemized: false,
+                });
+            } else {
+                panic!("Quantifier {} does not have an annotation", term);
```
benjaminfjones, line 525: this is a DoS on certain valid SMT inputs :) Can we capture in an issue, or a TODO for better error handling?

src/cadical_propagator.rs
```diff
@@ -511,16 +352,16 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
 
                 // todo: have a helper function for this, because it gets included twice
                 // for literal in literals {
-                //     // if self.egraph.nelson_oppen_literals.contains(&literal) {
+                //     // if self.solver_state.egraph.nelson_oppen_literals.contains(&literal) {
```
benjaminfjones: Can we just remove all this commented out code?

src/egraphs/basic/egraph.rs
```diff
@@ -0,0 +1,2070 @@
...
+    /// Register a single term in the egraph (non-recursive).
+    /// Sets up terms_list, proof_forest, predecessors, function_maps for this term.
+    /// Register a single term (non-recursive). Children must already be registered.
+    /// Sets up terms_list, proof_forest, function_maps, and adds this term as a
+    /// predecessor of each of its children.
+    /// If `dynamic` is true, calls find_and_union_to_eclass to merge with any
+    /// existing congruent term (needed for quantifier instantiation and datatype axioms).
+    /// Returns true if the term was already registered.
+    /// Register a single term (non-recursive). Children must already be registered.
+    /// Takes raw IDs — no dependency on Term representation.
```
benjaminfjones: docstring is really confusing; it looks like it's triplicated (three different versions of a top-level doc statement?)

src/egraphs/basic/egraph.rs
```diff
@@ -0,0 +1,2070 @@
...
+                }
+            }
+        }
+
+        if y_parent == x {
+            return (true, highest_level_y, highest_hash_y);
+        };
+
+        // if they are in the same tree need to recompute the root
+        // this is super gnarly -> make better
```
benjaminfjones, line 555: haha :)

src/egraphs/basic/egraph.rs
```diff
@@ -0,0 +1,2070 @@
...
+            1,
+            "We are in check_self_disequality with t {}",
+            self.display_term(t)
+        );
+        let t_disequalities = &self.proof_forest[t as usize].disequalities();
+        debug_println!(19, 2, "We have t_disequalities {:?}", t_disequalities);
+
+        // TODO: should not need to sort disequalities here if we are using a deterministic hashmap
+        let sorted_disequalities: Vec<_> = t_disequalities.iter().collect();
+        // sorted_disequalities.sort_by_key(|(key, _)| **key);
```
benjaminfjones: rm

src/egraphs/basic/egraph.rs
```diff
@@ -0,0 +1,2070 @@
...
+                    self.display_term(*key),
+                    key,
+                    disequality,
+                    self.display_term(root),
+                    self.display_term(t),
+                    self.display_term(disequality.term),
+                    self.display_term(disequality.original_disequality.0),
+                    self.display_term(disequality.original_disequality.1)
+                );
+                // assert! ((smaller_term == self.find(disequality.original_disequality.0) && larger_term == self.find(disequality.original_disequality.1)) || (smaller_term == self.find(disequality.original_disequality.1) && larger_term == self.find(disequality.original_disequality.0)));
```
benjaminfjones: maybe better to make a comment here about what the expected invariant is at this program point

src/main.rs
```diff
@@ -131,23 +124,29 @@ fn main() -> Result<(), String> {
...
     // somewhat inefficient especially since we have to clone nnf_term, but I couldn't come up with a
     // better way to do this
     for nnf_term in nnf_terms {
-        let additional_constraints = check_for_function_bool(&nnf_term, &mut egraph, false);
+        let additional_constraints = check_for_function_bool(
+            &nnf_term,
+            &mut solver_state,
+            false,
+            args.ddsmt,
+            args.lazy_dt,
```
benjaminfjones, line 148: I'm not sure if `check_for_function_bool` is pre-existing, but the pattern of passing three booleans positionally is bug prone. I think `find_datatype_axioms` and `learn_ctor_selector_clauses` also suffer this

src/egraphs/traits.rs
```diff
@@ -0,0 +1,181 @@
...
+use std::hash::Hash;
+
+/// A SAT literal: positive means the variable is true, negative means false.
+/// Zero means "no decision" when returned from decision methods.
+pub type Lit = i32;
+
+/// Conflict explanation: the equalities that were asserted (and their
+/// congruence consequences) that together violate a disequality.
+#[derive(Debug, Clone)]
+pub struct Conflict<T> {
```
benjaminfjones, line 21: I guess `T` is the type for terms? Can you document what it's supposed to represent?

src/egraphs/basic/egraph.rs
amarshah1: I think we can get away with not using `register_true` and register_false`. I will make the change

src/egraphs/traits.rs
amarshah1: Yeah its dead right now, but it will be useful for future optimizations

src/arithmetic/lp.rs
amarshah1: Yeah previously I didn't use `find` here. I think this might be a bug but didn't find a case to trigger it. I kept it as is for the sake of preserving the logical structure in this PR, but I will capture this as an issue.

amarshah1: See #120

src/egraphs/traits.rs
amarshah1: I will add a comment expressing this

src/solver_state.rs
amarshah1: Ok done, see #121

src/cadical_propagator.rs
amarshah1: done!

src/egraphs/basic/egraph.rs
amarshah1: Sorry I fixed it! This was a Claude generated docstring that I missed when I was going over everything :(

src/egraphs/basic/egraph.rs
amarshah1: done!

src/egraphs/basic/egraph.rs
amarshah1: done!

HuStmpHrrr:
> We regress from 66.1% to 65.5% on the Verus benchmarks. I anticipate this is largely because we waste time converting from the Yaspar representation to the egraph representation

any chance we can abstract yaspar ASTs with traits as well? Really it is possible to define entirely abstract interactions among only traits without defining any concrete data representations.

amarshah1:
> any chance we can abstract yaspar ASTs with traits as well? Really it is possible to define entirely abstract interactions among only traits without defining any concrete data representations.

Yes we could and I suggested it initially, but Bruno was against it. We should be able to have a faster congruence closure implementation using our own internal representation

src/main.rs
amarshah1: Yeah this carried over from the implementation before. I will capture this as an issue #122

src/egraphs/traits.rs
amarshah1: added!

src/arithmetic/lp.rs
benjaminfjones: Thanks! I'll also give some thought to whether there's a bug here or not.

benjaminfjones (approved):

amarshah1:
We solve 68.6% of benchmarks after fixing the issue with `solver_walk_term` (as opposed to 68.8% before we made any changes). So it seems like modularizing has a negligible effect on performance :)

benjaminfjones:
> We solve 68.6% of benchmarks after fixing the issue with `solver_walk_term` (as opposed to 68.8% before we made any changes). So it seems like modularizing has a negligible effect on performance :)

Nice!

Issue #50, opened 2026-03-04
Support inductive datatypes.


amarshah1:
We should add support for inductive datatypes. The high-level idea behind this is to add a [lazy "occurs check",](https://z3prover.github.io/papers/z3internals.html#sec-algebraic-datatypes) where you check for cycles in all inductive datatype terms. Note that this can be done in `cb_check_found_model` in `cadical_propagator` (when we do `check_integer_constraints_satisfiable` and `instantiate_quantifiers`).

Some things to be careful about when you do this:

- Make sure you don't regress on benchmarks where there are no inductive datatypes. You can check for inductive datatypes at the start.
- Makes sure to handle parametric inductive datatypes correctly.
- Be careful about the correct heuristics/ordering with `check_integer_constraints_satisfiable` and `instantiate_quantifiers`. This can matter a lot for performance

PR #126, opened 2026-06-18
Adding support for inductive datatypes

amarshah1:
*Issue #, if available:* #50

*Description of changes:* I have added support for inductive datatypes. This requires a few changes:

1. Instead of panicking when we detect an inductive datatype, we detect inductive datatypes and update `recursive_args`, `base_constructors`, and `recursive_sorts`.
2. We defer adding the only one tester check `isC1(t) \/ ... \/ isCk(t)` for inductive datatypes until the end. This matches the default z3 heuristic and with our current implementation, we could end up in an infinite loop if we dont do this.
3. We add decisions to the solver to force each datatype to pick the base-case selector. This can improve performance and avoids infinite loops
4. For terms of the form `C(t1,...,tk)`, we need to eagerly add the axiom `is-C(C(t1,...,tk))` since otherwise the constructors can get added out of order which causes issues
5. We add a [lazy occurs check](https://z3prover.github.io/papers/z3internals.html#sec-algebraic-datatypes) to test for well-foundedness

This can probably be optimized much further and requires more detailed benchmarking. I have captured this in issue #130.

I used Claude while approving changes myself and rewriting some parts. I also tried to use Claude fully autonomously last night to fix a bug. When, I came back this morning, it had an incorrect solution, but I went back and forth with it a few times and it figure it out. I am not sure if the time it spent last night was actually helpful


amarshah1:
Out of the 17 Verus benchmarks that use inductive datatypes, we solve 7 of them. In total we solve 68.9% on Verus benchmarks (previously 68.6%). I think this improvement is caused by the fact that we solve some of the Verus benchmarks that were previously failing

src/cadical_propagator.rs
```diff
@@ -410,7 +434,21 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
...
+
+        // For recursive datatypes, prefer base-case constructors to avoid infinite expansion
+        if self.solver_state.datatype_info.has_recursive_datatype() {
+            for &lit in &self.solver_state.base_case_tester_lits {
+                let idx = lit.unsigned_abs() as usize;
+                if idx >= self.assignments.len() {
+                    self.assignments.resize(self.assignments.len() * 2, 0);
+                }
+                if self.assignments[idx] == 0 {
+                    eprintln!("    cb_decide -> {}", lit);
```
benjaminfjones: looks like a stray debug statement

benjaminfjones:
Looks like all the e-graph trait commits and Caden's eDRAT changes also got included in this one

amarshah1:
> Looks like all the e-graph trait commits and Caden's eDRAT changes also got included in this one

Yeah its cause I did this before I merged those into main. I need to rebase

src/cadical_propagator.rs
amarshah1: fixed!

src/cadical_propagator.rs
```diff
@@ -403,7 +427,20 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
...
         debug_println!(7, 0, "PROPAGATOR: Decision callback invoked");
-        // always no decision
+
+        // For recursive datatypes, prefer base-case constructors to avoid infinite expansion
+        if self.solver_state.datatype_info.has_recursive_datatype() {
+            for &lit in &self.solver_state.base_case_tester_lits {
+                let idx = lit.unsigned_abs() as usize;
+                if idx >= self.assignments.len() {
+                    self.assignments.resize(self.assignments.len() * 2, 0);
+                }
```
benjaminfjones: It looks like there's a latent panic here when `idx` is large. Why not do repeated doubling in a while loop until you know for sure `self.assignments` is big enough?

benjaminfjones: like at line 104 in this same file

src/datatypes/occurs_check.rs
```diff
@@ -0,0 +1,186 @@
...
+    let mut clause: Vec<i32> = Vec::new();
+    let n = cycle_edges.len();
+
+    for i in 0..n {
+        let this_child = cycle_edges[i].1;
+        let next_parent = cycle_edges[(i + 1) % n].0;
+
+        if this_child == next_parent {
+            continue;
+        }
```
benjaminfjones: suppose this is always true in the loop, then we return the empty clause. Is that OK? It seems like it will force UNSAT unconditionally

src/solver_state.rs
benjaminfjones: Is this dead code now that it's call was removed from `main`?

src/main.rs
```diff
@@ -76,14 +76,6 @@ fn main() -> Result<(), String> {
 
     solver_state.register_bool_constants(&true_term, &false_term);
 
-    // checking if we have any inductive datatypes -> if we do panic!
-    // gets the info about our datatypes
-    if let Some(dt) = solver_state.check_for_recursive_datatypes() {
```
benjaminfjones, line 81: See comment about check_for_recursive_datatypes being dead

src/solver_state.rs
```diff
@@ -142,6 +142,9 @@ pub struct SolverState {
 
     /// Whether to skolemize eagerly.
     pub eager_skolem: bool,
+
+    /// SAT literals for base-case constructor testers (used by cb_decide to prefer base cases)
+    pub base_case_tester_lits: Vec<i32>,
```
benjaminfjones, line 147: does this ever get cleared or deduped? I don't see it happening anywhere. It seems that `cb_decide` scans over it so I was wondering if that will become progressively slower and slower as the run continues and backtracking/restarting happens

src/cadical_propagator.rs
```diff
@@ -403,7 +427,20 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
 
     fn cb_decide(&mut self) -> i32 {
         debug_println!(7, 0, "PROPAGATOR: Decision callback invoked");
-        // always no decision
+
+        // For recursive datatypes, prefer base-case constructors to avoid infinite expansion
+        if self.solver_state.datatype_info.has_recursive_datatype() {
+            for &lit in &self.solver_state.base_case_tester_lits {
```
benjaminfjones, line 433: see comment about `base_case_tester_lits` being cleaned up upon backtracking

src/datatypes/axioms.rs
```diff
@@ -286,6 +347,39 @@ pub fn learn_ctor_selector_clauses(
     vector
 }
 
+/// For a term of the form f(t1, ..., tk) where f is a constructor, assert is-f(term) as a unit clause.
+///
+/// This is needed because tester clauses are deferred for recursive sorts (to avoid infinite
```
benjaminfjones, line 352: Thanks for the great comments!

src/cadical_propagator.rs
amarshah1: yes, good catch!

src/datatypes/occurs_check.rs
amarshah1: This can actually never be true, so I will just delete it

src/solver_state.rs
amarshah1: yes, I will remove it

src/main.rs
amarshah1: done!

src/solver_state.rs
amarshah1: I think these are alright. We introduce `base_case_tester_lits` lazily, to avoid the number growing too large. If Cadical learns a conflict that prevents one of these decisions, it can ignore it. It seems to perform alright when tested so far. I think the optimal solution would be to use a phase instead of forcing decisions (see #130), but this is out of the scope for this PR.

src/cadical_propagator.rs
amarshah1: See response to the other comment

src/solver_state.rs
amarshah1: Essentially, we are using this to say that we want to pick the base case unless we absolutely have no other choice

src/datatypes/occurs_check.rs
benjaminfjones: Oh, got it. That makes sense.

src/solver_state.rs
benjaminfjones: Sounds good!

benjaminfjones (approved):
Hurray for inductive datatypes!

PR #128, opened 2026-06-19
Disable the default branch-and-bound step limit

benjaminfjones:
*Description of changes:*

This change improves the stability of Sundance, especially after several open PRs are applied, e.g. #99 and #114. Without this (or similar change), those PRs can't go through.

What we observe is that if the branch and bound limit is in place, certain regression tests can go from timeout to unknown depending on the platform (macos vs. Linux), especially on the CI workers.

Disabling the limit is a temporary measure. A better alternative will be to investigate if arithmetic gives us unknown and there are no quantifier instantiations to be found, only then do we run arithmetic with no resource limitations.

----

### DIFF
--- a/src/arithmetic/lia/config.rs
+++ b/src/arithmetic/lia/config.rs
@@ -19,10 +19,7 @@ impl Default for SolverConfig {
     fn default() -> Self {
         Self {
             tableau_kind: TableauKind::Dense,
-            // 2^17; high enough that regression tests pass as expected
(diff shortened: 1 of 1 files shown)


amarshah1 (approved):
This looks good. I think it makes sense to eventually have a limit, but you just do one final check of arithmetic with no limit before returning unknown. I captured this as a subissue: #129

HuStmpHrrr (approved):

PR #131, opened 2026-06-23
Adding a signature table to the egraph

amarshah1:
*Issue #, if available:* None

*Description of changes:* We originally used `union_predecessors` to handle unions between predecessors in the egraph which computed the canonical representation of each of the predecessors of two terms and then iterate over them. Instead, we maintain a signature table of all of the canonical terms. This is much simpler and leads to a significant speedup :)


amarshah1:
This improves performance from 68.6% on Verus benchmarks to 71.4%, while also making the code simpler!!

src/egraphs/basic/egraph.rs
```diff
@@ -485,147 +462,6 @@ impl Egraph {
...
-                }
-            }
-        }
-
-        if y_parent == x {
-            return (true, highest_level_y, highest_hash_y);
-        };
-
-        // if they are in the same tree need to recompute the root
-        // this is super gnarly -> make better
```
benjaminfjones, line 555: yay! :)

src/egraphs/basic/egraph.rs
```diff
@@ -485,147 +462,6 @@ impl Egraph {
         }
     }
 
-    // checks if x and y are equal in union find datastructure
-    // if they are equal, returns the largest level in both their paths to a
-    // common ancestor and the corresponding hash
-    fn check_equal(&self, x: u32, y: u32) -> (bool, usize, u32) {
```
benjaminfjones, line 491: great improvement to be able to remove this whole method

src/egraphs/basic/egraph.rs
```diff
@@ -345,9 +360,19 @@ impl Egraph {
             }
         }
 
-        // If dynamic, find and merge with existing congruent terms
+        // Insert into sig_table; if dynamic and sig already exists, merge via congruence.
+        if let Some(sig) = self.compute_signature(id) {
+            if let Some(&existing) = self.sig_table.get(&sig) {
+                if dynamic && self.find(existing) != self.find(id) {
+                    // TODO: propagate conflict — requires register_term_internal to return EgraphResult
```
benjaminfjones, line 367: Is this worth a tracking issue? I can't tell how big a deal it is. Is the conflict you're referring to happen when merging a term but that conflicts with an existing disequality set?

src/egraphs/basic/egraph.rs
```diff
@@ -210,6 +218,11 @@ pub struct Egraph {
     predecessors_created_by_quantifiers: DeterministicHashMap<u32, DeterministicHashSet<u32>>,
     /// if a quantifier instantiates (f t) and t = s, then we want to add (f.uid(), "f", [t.uid()])
     union_to_eclass: DeterministicHashSet<(u32, String, Vec<u32>)>,
```
benjaminfjones: The entries here are destructured in a few places and the last two expressions are ignored. Can it be changes to just a `DeterministicHashSet<u32>` ?

src/egraphs/basic/egraph.rs
```diff
@@ -1403,277 +1315,37 @@ impl Egraph {
...
-                11,
-                0,
-                "We are are adding predecessor {} (of  {}) to {} [level: {}, hash: {}]",
-                self.display_term(*pred_key),
-                self.display_term(pred_val.inner_term),
-                self.display_term(u),
-                level,
-                self.predecessor_hash
-            );
+        // Phase 3: Reinsert y_root's predecessors into sig_table with new canonical forms
```
benjaminfjones, line 1316: I can't find phases 1 & 2, but maybe the diff is just hard to read here

src/egraphs/basic/egraph.rs
```diff
@@ -728,30 +564,66 @@ impl Egraph {
...
+    /// Returns None for terms that don't participate in congruence (non-App/Eq/Ite ops, or opaque terms).
+    fn compute_signature(&self, term_id: u32) -> Option<SigKey> {
+        let entry = match &self.terms[term_id as usize] {
             TermSlot::Term(e) => e,
             _ => return None,
         };
-
-        let original_subterms = entry.children.as_slice().to_vec();
         let op = match &entry.op {
             Op::App(s) => CanonicalOp::App(s.to_string()),
```
benjaminfjones, line 574: Since this allocates on every call to `compute_signature`, it might be worth interning the strings at some point. On top of that, sig_table_insert/sig_table_remove `clone()` the full `SigKey` (String + possibly heap Children::ArityN)

benjaminfjones (approved):

src/egraphs/basic/egraph.rs
amarshah1: Yeah I think it's ok. As far as I can tell, it won't cause an error, since the conflict will be learnt in the next step. It is somewhat of a weird design choice

src/egraphs/basic/egraph.rs
amarshah1: Yes it can, great suggestion!

src/egraphs/basic/egraph.rs
amarshah1: yeah added!

src/egraphs/basic/egraph.rs
amarshah1: Yeah I agree, but I think the best way to do this would be to have a `register_function` function in the Egraph trait. Thus, the egraph can be agnostic towards the external representation of function names. I will capture this as a separate issue #135

Issue #117, opened 2026-06-09
Need to support Integer division/mod


amarshah1:
For instance, we return `sat` on this query

```
(declare-fun C () Int)
(assert (= C 1))
(assert (= 1 (div 0 C)))
(check-sat)
```

Z3 handles integer div/mod using the axiom: `∀x,y. y ≠ 0 → x = y * (div x y) + (mod x y)`

amarshah1:
There seems to be 3 errors to fix here:

1. When the solver calls the arithmetic solver, it needs to also provide information about div and mod

2. The solver needs to add the axioms `∀x,y. y ≠ 0 → x = y * (div x y) + (mod x y)` for each `div(a, b)` or `mod(a, b)` term that occurs.

3. When there is a nonlinear division, Sundance should complain. Note that if `div` occurs in the body of a quantifier, it could create instantiations that are not linear. It is unclear if you need to reason about non-linearity for the Verus queries

benjaminfjones:
Note: the query is still reported as SAT after https://github.com/yaspar-org/Sundance-SMT/pull/132 because `div` terms aren't being sent to the arithmetic solver.

PR #132, opened 2026-06-23
Fix integer division semantics in the arithmetic solver

benjaminfjones:
*Description of changes:*

Integer division (by a constant) was being treated the same as real division. This change fixes the bug by introducing new constraints based on Euclidean division semantics. 

New unit tests are provided that cover both upper and lower remainder bounds in the semantics.

----

### DIFF
--- a/src/arithmetic/lia/frontend.rs
+++ b/src/arithmetic/lia/frontend.rs
@@ -226,7 +226,7 @@ fn convert_linear_term(ctx: &mut ConvContext, term: &Term) -> FrontendResult<Lin
                         ))
                     }
                 }
-                REAL_DIV_SYMBOL | INT_DIV_SYMBOL => {
+                REAL_DIV_SYMBOL => {
                     debug_assert!(args.len() == 2, "chainable division is not supported");
                     let e1 = convert_linear_term(ctx, &args[0])?;
                     let e2 = convert_linear_term(ctx, &args[1])?;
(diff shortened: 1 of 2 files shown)


src/arithmetic/lia/solver_result_api.rs
```diff
@@ -31,17 +31,18 @@ impl SolverDecisionApi {
...
+                        var_term_map.get(&var).map(|term| (term.clone(), value))
+                    })
                     .collect();
                 Ok(SolverDecisionApi::FEASIBLE(Assignment::new(term_assign)))
             }
             SolverDecision::INFEASIBLE(conflict) => {
                 let term_conflict: Conflict<Term> = conflict
                     .iter()
-                    .map(|v| var_term_map.get(v).unwrap().clone())
+                    .filter_map(|v| var_term_map.get(v).cloned())
```
amarshah1, line 45: why do we need a filter_map instead of a map here?

src/arithmetic/lia/frontend.rs
```diff
@@ -1067,10 +1111,99 @@ mod tests {
...
+    /// (not 0, which exact division toward zero would give, and not -0.5 which
+    /// the pre-fix real-division handling produced).
+    #[test]
+    fn test_int_div_negative_numerator_floors() {
+        let smt_input = r#"
+(set-logic QF_LIA)
+(declare-const r Int)
+(assert (= r (div (- 1) 2)))  ; Euclidean: -1 = 2*(-1) + 1, so r = -1
+(assert (= r (- 1)))
+(check-sat)
```
amarshah1, line 1221: can you add one more regression test for when we have a negative divisor?

src/arithmetic/lia/frontend.rs
```diff
@@ -239,6 +239,60 @@ fn convert_linear_term(ctx: &mut ConvContext, term: &Term) -> FrontendResult<Lin
...
                     }
                 }
+                INT_DIV_SYMBOL => {
+                    // SMT-LIB integer division (Euclidean): (div e n) = q
+                    // where e = n * q + r, 0 <= r < |n|
+                    debug_assert!(args.len() == 2, "chainable division is not supported");
+                    let e1 = convert_linear_term(ctx, &args[0])?;
+                    let e2 = convert_linear_term(ctx, &args[1])?;
+                    if let Some(n) = e2.is_constant() {
+                        if n.is_zero() {
```
amarshah1, line 249: I think SMT-Lib defines division by zero as an uninterpreted function. At least thats how Z3 behaves. Not sure how we want to tackle these in Sundance

amarshah1:
This addresses some of the issues described in #117 , but not everything because there are some fixes that need to be made outside of the arithmetic solver. But pretty good overall!

amarshah1:
Do we have a similar issue with Integer Mod?

amarshah1:
> Do we have a similar issue with Integer Mod?

It seems like the internal solver doesnt handle mod at all? I think that is out of scope of this PR, but we should capture this as an issue

src/arithmetic/lia/frontend.rs
amarshah1: I don't think this needs to be addressed in this PR necessarily if its too complicated, but it should be captured as an issue

amarshah1 (approved):
Looks good overall. Can you capture the divide by 0 and the integer mod things as two separate issues if you think they are problems that we need to address and are out of scope for this PR?

benjaminfjones:
> > Do we have a similar issue with Integer Mod?
> 
> It seems like the internal solver doesnt handle mod at all? I think that is out of scope of this PR, but we should capture this as an issue

Yeah, the arithmetic solver doesn't handle `mod` at all. I'll make a sub-issue.

src/arithmetic/lia/frontend.rs
benjaminfjones: I'll create an issue; it's definitely a gap

src/arithmetic/lia/solver_result_api.rs
benjaminfjones: Sorry, this was unrelated to the change at hand. I thought it would be cleaner to `filter_map` the `get` rather than map get and then unwrap everything. It's an invariant that every var in the conflict is in the `var_term_map` so the result is always the same either way.

src/arithmetic/lia/frontend.rs
benjaminfjones: https://github.com/yaspar-org/Sundance-SMT/issues/133

benjaminfjones:
> > > Do we have a similar issue with Integer Mod?
> > 
> > 
> > It seems like the internal solver doesnt handle mod at all? I think that is out of scope of this PR, but we should capture this as an issue
> 
> Yeah, the arithmetic solver doesn't handle `mod` at all. I'll make a sub-issue.

Issue to support `mod` in arithmetic https://github.com/yaspar-org/Sundance-SMT/issues/134#issue-4731003241

src/arithmetic/lia/frontend.rs
benjaminfjones: good idea

benjaminfjones: done!

benjaminfjones: This is a good example because it illustrates how flooring properly changes  direction when the denominator changes sign.

PR #137, opened 2026-06-25
Batching quantifier instantiations

amarshah1:
*Issue #, if available:*

*Description of changes:* We modify our quantifier instantiation loop, so we first compute all of the instantiations before we make the substitutions. Before when we were looping through quantifiers: $Q_1, Q_2, Q_3, \ldots$, when trying to e-match against $Q_2$, this would include all of the terms that were created by instantiating $Q_1$. This led to instability (as noticed by Long) since the order of quantifiers could have a large affect on performance.

This also shows a performance improvement on the Verus benchmarks from 71.4% to 72.4%. I think this points to a bigger issue, that we are learning quantifier instantiations too eagerly, but I am not sure how to fix it yet.


amarshah1:
This seems to have a positive effect on performance on Verus benchmarks. We are going from 71.4% to 72.4%

benjaminfjones (approved):

PR #138, opened 2026-06-26
Making constants root of their eclass

amarshah1:
*Issue #, if available:*

*Description of changes:* This modifies the egraph, so that constants such as `true`, `false`, `1`, `2`, ... are always the root of their eclass and will immediately detect a conflict if they are set equal to each other

### DIFF
--- a/src/egraphs/basic/egraph.rs
+++ b/src/egraphs/basic/egraph.rs
@@ -899,7 +899,7 @@ impl Egraph {
             return EgraphResult::with_conflict(Conflict {
                 equalities,
                 disequality: (t1, t2),
-                diseq_lit,
+                diseq_lit: None,
             });
         }
         let hash = self.predecessor_hash;
@@ -1175,7 +1175,7 @@ impl Egraph {
             return EgraphResult::with_conflict(Conflict {
                 equalities,
                 disequality: (x_root, y_root),
-                diseq_lit: 0,
+                diseq_lit: None,
             });
         }
 
@@ -1236,7 +1236,7 @@ impl Egraph {
                 return EgraphResult::with_conflict(Conflict {
                     equalities,
                     disequality: disequality.original_disequality,
-                    diseq_lit: disequality.diseq_lit,
+                    diseq_lit: Some(disequality.diseq_lit),
                 });
             } else {
                 panic!(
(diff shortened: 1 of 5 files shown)


amarshah1:
This improves performance from 71.4% to 80.0% on Verus benchmarks which is huge, but a little surprising. I'm working on running this with the new Valido

src/solver_state.rs
```diff
@@ -799,49 +812,11 @@ pub fn process_assignment(
...
     };
 
-    debug_println!(
-        4,
-        0,
-        "We are in process_assignment, checking for contradiction with true and false",
+    debug_assert!(
+        solver_state.egraph.find(true_egraph_id) != solver_state.egraph.find(false_egraph_id),
+        "true and false merged without conflict being detected"
     );
```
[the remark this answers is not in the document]
benjaminfjones: Normally, I like debug asserts better, but in our case we don't have good ways of exercising them. The regression tests require a release build and otherwise we just have the unit tests which of which there are not many for the core solver.

src/solver_state.rs
benjaminfjones: Should this also be guarded by `conflict.diseq_lit != 0` for consistency?

benjaminfjones: also, lifting `0` to a named constant throughout the codebase would be a nice code documentation improvement

src/solver_state.rs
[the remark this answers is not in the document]
amarshah1: Yeah I think this is fine here because this is just a sanity check. We can run on the regression tests with debug_assert every now and then. We could also change the regression tests to not use the release version

src/solver_state.rs
amarshah1: Its not necessary since the diseq_lit as 0 can only be assigned in the assert_equal path

amarshah1: I will make it an option. I think this is an easier way to avoid confusion

src/egraphs/basic/egraph.rs
```diff
@@ -1144,6 +1154,56 @@ impl Egraph {
             return EgraphResult::ok();
         }
 
+        let x_root_is_const = self.is_constant(x_root);
+        let y_root_is_const = self.is_constant(y_root);
+
+        if x_root_is_const && y_root_is_const {
+            let mut equalities = Vec::new();
```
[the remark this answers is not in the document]
amarshah1: They can never be equal. Can add a debug_assert there

amarshah1:
Can get up to 80.9% on Verus benchmarks when combined with the recently merged in optimizations!

benjaminfjones (approved):

PR #139, opened 2026-06-26
Replace term vec clones with iteration

leeabarnett:
*Issue #, if available:*

*Description of changes:* I was looking back at this code and realized we were doing an unnecessary clone. Minor improvement but somewhat noticeable. The plot is for `--arithmetic internal`; I didn't benchmark the z3 path.

 Screenshot 2026-06-26 at 9 21 57 AM
---


amarshah1 (approved):
This looks great, thanks!

PR #140, opened 2026-06-29
Lazy quantifier instantiation

amarshah1:
*Issue #, if available:*

*Description of changes:* This implements lazy quantifier instantiation, where instead of learning a batch of quantifier instantiations at the same time, we learn quantifier instantiations one-at-a-time. However, if we did this naively, cadical would require a backtrack after every quantifier instantiation. Instead, we have to use the an experimental cadical feature "elevate", which exists in a branch. I created my own fork of cadical-sys that uses this branch and Sundance will have to use this for right now

This provides a *very* modest performance improvement, but will be super useful for adding other performance improvements on top of it, like term garbage collection and heuristics for exploring quantifier instantiation order. 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ local-z3 = ["z3-solver"]
 [dependencies]
 yaspar-ir = "=2.7.2"
 sat-interface = "0.1.1"
-cadical-sys = { path = "cadical-sys-local" }
+cadical-sys = { git = "https://github.com/amarshah1/cadical-sys.git", branch = "elevate" }
 
 array2d = "^0.3"
 clap = { version = "4.5.40", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ local-z3 = ["z3-solver"]
 [dependencies]
 yaspar-ir = "=2.7.2"
 sat-interface = "0.1.1"
-cadical-sys = { version = "0.7.0" }
+cadical-sys = { path = "cadical-sys-local" }
 
 array2d = "^0.3"
 clap = { version = "4.5.40", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,8 @@ local-z3 = ["z3-solver"]
 [dependencies]
 yaspar-ir = "=2.7.2"
 sat-interface = "0.1.1"
-cadical-sys = { git = "https://github.com/amarshah1/cadical-sys.git", branch = "elevate" }
+# Pin the elevate fork until upstream cadical-sys exposes CaDiCaL's elevate option.
+cadical-sys = { git = "https://github.com/amarshah1/cadical-sys.git", rev = "2f234ec95dd36ca74057691cf2fd4e8766a3c40f" }
 
 array2d = "^0.3"
 clap = { version = "4.5.40", features = ["derive"] }
--- a/src/cadical_propagator.rs
+++ b/src/cadical_propagator.rs
@@ -280,39 +280,39 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
         if let Some(mut pending) = self.pending.take()
             && let Some(instances) =
                 materialize_next(&mut pending, self.solver_state, &self.proof_tracer)
-            {
-                for inst in &instances {
-                    match inst {
-                        Instantiation { clauses } => {
-                            for clause in clauses {
-                                for lit in clause {
-                                    self.add_observed_variable(*lit);
-                                    self.add_lit_to_proof_tracer(*lit);
-                                }
-                                self.disequalities.borrow_mut().push(clause.clone());
+        {
+            for inst in &instances {
+                match inst {
+                    Instantiation { clauses } => {
+                        for clause in clauses {
+                            for lit in clause {
+                                self.add_observed_variable(*lit);
+                                self.add_lit_to_proof_tracer(*lit);
                             }
-                            self.stats.instantiations += 1;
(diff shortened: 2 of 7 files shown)


amarshah1:
Before we merged in the egraph constant change, we were seeing an improvement from 72.4% to 77.0%. However, after the egraph constant merge we only see an improvement from 80.9% to 81.1%. Here new_default is Sundance after we merge in egraph constant change and lazy QI is this PR

new_default_vs_lazyQI

amarshah1:
I spent some time diving into what causes some of these regressions. I believe since we are in a slightly different cadical version now, the SAT solver behaves slightly differently and this can help us some times and hurt us some times. Two queries I looked at:

1. splinterdb/journal__PagedJournalRefinement_v.18.smt2

Solved in 1.78s on baseline, 28-55s with lazy QI + elevate=2/3. The solver needs all ~395 instantiations across 5 rounds. I initially thought we were seeing a regression because you have to go pretty deep in the search, but that does not seem to be the case. Finding instantiation is fast (~2.3s), but with elevate=2/3, CaDiCaL's post-instantiation search takes 25+ seconds. This can be fixed by using elevate=1 (brings it to ~4-9s). But setting elevate to -1 or 1 does not make a performance difference overall

2. verismo/tspec__math__bits_p.1.smt2

Solved in 0.2s on standard CaDiCaL 2.1.3, times out on rel-2.1.3-elevate-new regardless of elevate setting. Never reaches quantifier instantiation — gets stuck in arithmetic/Nelson-Oppen conflicts. The SAT solver repeatedly tries to make 0 = (1<<N) or (< 0 (1<<N)) AND (> 0 (1<<N)) true for 64 bit positions, the arithmetic solver rejects each one. Claude thinks it is aused by a change in propagate.cpp where the elevate branch always uses assignment_level for propagations (removing the opts.chrono guard), which changes CaDiCaL's search behavior on this type of problem independently of the elevate option, but I have not verified this at all

src/egraphs/basic/egraph.rs
```diff
@@ -1640,7 +1634,7 @@ fn valid_hash(hash: u32, level: usize, predecessor_level: &[u32]) -> bool {
         hash,
         level
     );
-    hash >= predecessor_level[level] || hash == 0 || level == 0 // todo: I added this level ==0 ~> I think this is correct but need to double check to be sure
+    hash >= predecessor_level[level] || level == 0
```
amarshah1, line 1637: checking hash == 0 here was actually a bug that I + Claude found while making the changes for this PR

src/quantifiers/quantifier.rs
```diff
@@ -20,36 +20,45 @@ use yaspar_ir::ast::{LetElim, Sort, Str, Substitute, Substitution, Term, TermAll
...
+    pub quantifier_id: u64,
 }
 
-struct DeferredSkolemization {
-    skolem: Term,
-    skolem_vars: Vec<(Str, Sort)>,
-    is_exists: bool,
-    literal: i32,
+pub struct DeferredSkolemization {
+    pub skolem: Term,
```
benjaminfjones: Would you mind documenting public structs, fields, and functions? It looks like a bunch of things turned public in this PR.

src/cadical_propagator.rs
```diff
@@ -275,6 +274,45 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
             return false;
         }
 
+        // If we have pending instantiations from a previous round, serve one immediately
```
benjaminfjones: what does "serve" mean here?

Cargo.toml
```diff
@@ -20,7 +20,7 @@ local-z3 = ["z3-solver"]
 [dependencies]
 yaspar-ir = "=2.7.2"
 sat-interface = "0.1.1"
-cadical-sys = { version = "0.7.0" }
+cadical-sys = { git = "https://github.com/amarshah1/cadical-sys.git", branch = "elevate" }
```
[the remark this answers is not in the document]
benjaminfjones: agree with this one; pinning a commit or version tag would be good

src/cadical_propagator.rs
amarshah1: basically just "instantiate", will reword

Cargo.toml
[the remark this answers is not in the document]
amarshah1: done!

src/quantifiers/quantifier.rs
```diff
@@ -152,28 +161,49 @@ pub fn instantiate_quantifiers(
...
+    proof_tracer: &Rc<RefCell<SMTProofTracer>>,
+) -> Option<Vec<QuantifierInstance>> {
+    let ddsmt = solver_state.ddsmt;
+    let lazy_dt = solver_state.lazy_dt;
+
+    // Skolemizations first
+    if let Some(deferred) = pending.deferred_skolemizations.pop() {
+        let results = process_deferred_skolemizations(
+            vec![deferred],
+            solver_state,
```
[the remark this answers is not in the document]
amarshah1: done!

src/quantifiers/quantifier.rs
amarshah1: Sorry, I made all of those private/only visible in the crate

src/cdcl.rs
```diff
@@ -42,6 +42,7 @@ pub fn cdcl_decision_procedure(
     timeout: u64,
 ) -> (Status, SolverStats) {
     let mut solver = CaDiCal::new();
+    solver.set("elevate".to_string(), 3);
```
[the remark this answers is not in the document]
amarshah1: done

benjaminfjones (approved):

Issue #110, opened 2026-05-24
Creating too many interface lemmas for Nelson-Oppen theory combination


amarshah1:
We are learning a bunch of Nelson-Oppen clauses of the form `x=y \/ x < y \/ x > y'.

I think we can avoid this using the strategy discussed in the [Yices 2.2 paper](https://yices.csl.sri.com/papers/cav2014.pdf). Essentially you need a Nelson-Oppen clause as above if LIA |= x = y and UF |= x != y. Right now we are creating a clause only if LIA |= x = y, but we should also check that UF |= x != y before adding such a clause. 

@BrunoDutertre let me know if this sounds good or if there is a better way to do Nelson-Oppen.

amarshah1:
Actually this is maybe not so simple, say you have the query

```
(declare-fun % () Int)
(declare-fun h (Int) Bool)
(declare-const k Int)
(assert (h k))
(assert (= k %))
(assert (not (h (+ 0 %))))
(check-sat)
```

Then we want to learn a interface lemma for `%` and `(+ 0 %)`, but how are we to be sure whether to add this interface lemma or not?

amarshah1:
The Yices paper actually will try to merge `x` and `y` in the egraph and see if that creates a conflict. Maybe we can do something like this instead

PR #146, opened 2026-07-08
Lazy nelson oppen (probe version)

amarshah1:
*Issue #, if available:* #110

*Description of changes:* We support nelson-oppen theory combination via interface lemmas. However, we aggressively introduce interface lemmas. For instance for if the arithmetic solver returns `sat` with a model `w=x=y=z=1`, we would learn the interface lemmas `w=x \/ w<x \/ w>x`, `w=y \/ w<y \/ w>y`, `w=z \/ w<z \/ w>z`.

However, we can make this more efficient by checking if these equalities violate the current state of the egraph, before learning an interference lemma

### DIFF
--- a/src/arithmetic/lialp.rs
+++ b/src/arithmetic/lialp.rs
@@ -17,7 +17,7 @@ use crate::debug_println;
 use crate::egraphs::EgraphTrait;
 use crate::solver_state::SolverState;
 use crate::utils::{DeterministicHashMap, DeterministicHashSet};
-use dashu::{Integer, Rational};
+use dashu::{Integer, Rational, integer::IBig};
 use std::collections::HashMap;
 
 pub fn check_integer_constraints_satisfiable_lia(
@@ -119,13 +119,12 @@ pub fn check_integer_constraints_satisfiable_lia(
             let stats = ret.stats;
             match ret.decision {
                 SolverDecision::FEASIBLE(assignment) => {
-                    let mut model_hashmap: DeterministicHashMap<i64, DeterministicHashSet<u64>> =
+                    let mut model_hashmap: DeterministicHashMap<IBig, DeterministicHashSet<u64>> =
                         DeterministicHashMap::new();
                     for (term_id, root_var) in &roots {
                         if let Some(value) = assignment.get(root_var) {
-                            let val_i64: i64 =
-                                value.to_int().value().try_into().unwrap_or(i64::MAX);
-                            model_hashmap.entry(val_i64).or_default().insert(*term_id);
+                            let val: IBig = value.to_int().value().clone();
+                            model_hashmap.entry(val).or_default().insert(*term_id);
                         }
                     }
                     ArithResult::Sat(model_hashmap, stats)
--- a/src/arithmetic/nelsonoppen.rs
+++ b/src/arithmetic/nelsonoppen.rs
@@ -80,6 +80,42 @@ pub fn nelson_oppen_clause(literal: i32, solver_state: &mut SolverState) -> Opti
 //     }
 // }
 
+/// Build the three sub-terms (lt, gt, eq) for the trichotomy on (x, y).
+/// Returns None if the trichotomy has already been emitted for this pair.
+/// Marks the pair as emitted on the first successful call.
+pub fn nelson_oppen_trichotomy_terms(
+    x: u64,
+    y: u64,
+    solver_state: &mut SolverState,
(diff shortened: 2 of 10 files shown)


src/cadical_propagator.rs
```diff
@@ -354,35 +381,94 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
...
+                            .egraph
+                            .assert_equal(x_e, y_e, attempt_level);
+                        if let Some(c) = result.conflict {
+                            // Undo just this conflicting merge; keep earlier merges.
+                            self.solver_state.egraph.backtrack_to(attempt_level - 1);
+                            next_probe_level = attempt_level;
+                            conflicts.push(c);
+                            if conflicts.len() >= self.max_arith_conflicts_per_round {
+                                break 'outer;
                             }
```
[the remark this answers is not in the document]
amarshah1: we will assert that the number of conflicts is >0

amarshah1:
With the internal arithmetic solver, we improve from 72.4% to 75.7%

Using the z3 arithmetic solver, we improve from 81.3% to 83.8%

image

image

src/arithmetic/lialp.rs
```diff
@@ -119,13 +119,12 @@ pub fn check_integer_constraints_satisfiable_lia(
...
-                    let mut model_hashmap: DeterministicHashMap<i64, DeterministicHashSet<u64>> =
+                    let mut model_hashmap: DeterministicHashMap<IBig, DeterministicHashSet<u64>> =
                         DeterministicHashMap::new();
                     for (term_id, root_var) in &roots {
                         if let Some(value) = assignment.get(root_var) {
-                            let val_i64: i64 =
-                                value.to_int().value().try_into().unwrap_or(i64::MAX);
-                            model_hashmap.entry(val_i64).or_default().insert(*term_id);
+                            let val: IBig = value.to_int().value().clone();
+                            model_hashmap.entry(val).or_default().insert(*term_id);
```
benjaminfjones, line 127: Nice! I forgot about this gap.

src/arithmetic/nelsonoppen.rs
```diff
@@ -2,86 +2,18 @@
...
-// }
-
-// learn the clause x = y \/ x > y \/ x < y
-pub fn nelson_oppen_clause_pair(x: u64, y: u64, solver_state: &mut SolverState) -> Option<Term> {
+/// Build the three sub-terms (lt, gt, eq) for the trichotomy on (x, y).
+/// Returns None if the trichotomy has already been emitted for this pair.
+/// Marks the pair as emitted on the first successful call.
+pub fn nelson_oppen_trichotomy_terms(
+    x: u64,
+    y: u64,
```
benjaminfjones, line 14: what do the u64's represent here? are they e-graph IDs?

amarshah1:
I also tried a version of this that did something like if x = y were equal in the arithmetic model, it tried to merge x and y in the egraph and if that edge was ever involved in a conflict, it would learn the trichotomy (see #142)

It was more complicated and potentially unsound. It got a similar speedup on the z3 branch, but it did get an even better speedup on the internal arithmetic branch

src/arithmetic/nelsonoppen.rs
amarshah1: u64 is the id in the solver state and u32 is the id in the egraph

amarshah1: I can use EgraphId, SolverStateId instead of u32, u64

benjaminfjones: nevermind, answered my own question below. It would be a nice code-documentation improvement at some point to newtype these, or make type aliases at least.

benjaminfjones: woops! comment race condition :) That sounds great

benjaminfjones:
Can you come up with any good regression tests for the new combination approach?

amarshah1:
> Can you come up with any good regression tests for the new combination approach?

Yup, can add a few. Its a little bit tricky since it's just a performance optimization, so there are no new features to test.

benjaminfjones:
> > Can you come up with any good regression tests for the new combination approach?
> 
> Yup, can add a few. Its a little bit tricky since it's just a performance optimization, so there are no new features to test.

Gotcha. That's okay then since all the existing regressions pass.

benjaminfjones (approved):

PR #149, opened 2026-07-09
Incremental arithmetic for z3 arith option

amarshah1:
*Issue #, if available:*

*Description of changes:*


amarshah1:
The version I showed at the Sundance meeting had a soundness bug, but we still show a nice improvement from before (from 83.8% to 87.0%)

image

tests/regression/smt_files/arithmetic/quant_pattern_z3_unknown_2.smt2
```diff
@@ -0,0 +1,49 @@
...
+(declare-fun v9 () Int)
+(declare-fun f0 (Int Int Int) Int)
+(declare-fun f1 (Int Int) Int)
+(declare-fun f2 (Int) Int)
+(declare-fun f3 (Int Int) Int)
+(declare-fun f4 (Int Int Int) Int)
+(declare-fun f5 (Int Int) Int)
+(declare-fun f6 (Int Int) Int)
+(assert (=> (= v1 v0) (= (f5 v1 0) (f5 v0 0))))
+(assert (and (= (- v0 (f3 3 (f5 v4 (- v7)))) v9) (= (+ v8 v6) (- (- (+ (- v3) (+ (- 10) v3))))) (= v4 3)))
```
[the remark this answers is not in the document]
benjaminfjones: `v4` is declared `Int`, so the first finding is not correct. I also don't see any use of unary comparison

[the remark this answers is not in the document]
benjaminfjones: cvc5, with strict parsing, passes on this file and reports unsat.

src/arithmetic/z3incremental.rs
```diff
@@ -0,0 +1,611 @@
...
+    /// Stack of push counts per decision level. Entry `i` is how many
+    /// `z3::push`es we've done at level `i`; on `notify_backtrack(level)`
+    /// we pop everything above `level`.
+    /// Indexed by decision level; grows as decision level grows.
+    push_counts: Vec<u32>,
+    /// Definitional pinnings (e.g. `var_5 == 5`, `var_{(+xy)} == var_x + var_y`).
+    /// These are theory facts that hold globally, so we replay them as
+    /// assumptions on every `check_assumptions` rather than asserting them into
+    /// a specific scope (where a later `pop` would erase them).
+    pinned_defs: Vec<Bool>,
```
benjaminfjones: Why not assert these first, before any `push` and then not have to replay them?

amarshah1: Yeah the problem is that these get added lazily, so they could be added at a higher level, so they would get backtracked on and then we would need to readd them. This a little more complicated, but might be slightly faster, so I can try it

tests/regression/smt_files/arithmetic/quant_pattern_z3_unknown_2.smt2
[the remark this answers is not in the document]
amarshah1: yeah I think copilot is tripping

src/arithmetic/z3incremental.rs
```diff
@@ -0,0 +1,611 @@
...
+            self.current_level
+        );
+    }
+
+    /// Drain the egraph's arithmetic merge queue. For each merge `(a, b)`,
+    /// allocate the SAT literal for `(= t_a t_b)` via `make_eq`, then push
+    /// `var_a == var_b` into Z3 via `assert_and_track` with a tracker keyed
+    /// on that literal. Result: each merge is a normal tracked atom, and
+    /// Z3's unsat core will blame exactly the merge lits that were actually
+    /// used in the conflict — no over-approximation.
```
benjaminfjones: I think this assumes the unsat assumption set is minimal? Doesn't matter for the implementation I think, just thinking out loud.

amarshah1: Yeah it does

src/arithmetic/z3incremental.rs
```diff
@@ -0,0 +1,611 @@
...
+                a,
+                b,
+                lit,
+                self.current_level
+            );
+        }
+        new_lits
+    }
+
+    /// Run `solver.check()`. On SAT, group current arithmetic-term roots by
```
benjaminfjones: Have you looked at how much grouping occurs in some of our benchmarks? I'm curious how large the classes tend to be. In synthetic examples there tend to be a lot of variables assigned 0, e.g. slack variables.

src/arithmetic/z3lazy.rs
```diff
@@ -0,0 +1,611 @@
...
+        }
+        new_lits
+    }
+
+    /// Run `solver.check()`. On SAT, group current arithmetic-term roots by
+    /// model value so the caller's model-based Nelson-Oppen probe can run.
+    /// On UNSAT, translate the unsat core back into SAT literals.
+    pub fn check(&mut self, solver_state: &mut SolverState) -> ArithResult {
+        debug_println!(21, 0, "[z3lazy] check() at level {}", self.current_level);
+        // Ensure every arithmetic-term class root has been introduced to Z3
```
benjaminfjones: Do you really mean "class" rather than "class root" here? Checking my understanding that it's the e-graph class being associated w/ z3 variable rather than the root specifically.

src/arithmetic/z3incremental.rs
```diff
@@ -0,0 +1,611 @@
...
+                }
+                // Merges are now tracked via `assert_and_track` (see
+                // `drain_merge_queue`), so Z3's unsat core cites exactly the
+                // merge lits (via `make_eq`-allocated SAT lits) that were
+                // actually used in the conflict. No over-approximation needed.
+                if lits.is_empty() {
+                    // Sanity guard: if no tracked lits appear in the core, the
+                    // contradiction lives entirely in pinned defs — a bug.
+                    // Fall back to blaming every active lit rather than
+                    // reporting unsound `Unsat([])`.
```
benjaminfjones: So should this actually be a panic if reaching it represents a bug? The comment is not reassuring.

src/arithmetic/z3incremental.rs
amarshah1: Yeah I'm not really sure. This is something worth investigating

src/arithmetic/z3lazy.rs
amarshah1: That comment was wrong sorry. It is per term not per class or class root

src/arithmetic/z3incremental.rs
amarshah1: Yeah you're right. I added a panic

amarshah1:
@HuStmpHrrr can you change the branch protection rules to accommodate the new workflows?

benjaminfjones (approved):

PR #151, opened 2026-07-13
[DRAFT] Factor incremental arithmetic into a trait

amarshah1:
*Description of changes:*

Factoring incremental arithmetic into a trait

----


src/arithmetic/incremental_solver/z3.rs
```diff
@@ -0,0 +1,254 @@
...
+                        let signed = if self.active_lits.contains(&abs_lit) {
+                            abs_lit
+                        } else {
+                            -abs_lit
+                        };
+                        Some(-signed)
+                    })
+                    .collect();
+                assert!(
+                    !lits.is_empty(),
```
amarshah1, line 288: We end up hitting this panic... But was not an issue before the egraph trait -> need to fix

amarshah1: We error on a ton of Verus benchmarks. Claude keeps trying to fix this by blaming all definitions, but this is ugly and also I think there might be something unsound going on, since it causes a huge performance improvemtn (90% success on Verus)

src/arithmetic/incremental_solver/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Push an equality `a == b` tracked by `lit`.
+    fn push_equality(&mut self, a: VarId, b: VarId, lit: i32);
+
+    /// Check satisfiability of all currently-pushed constraints + definitions.
+    /// On SAT, only vars registered with `report_in_model=true` appear in buckets.
+    fn check(&mut self) -> ArithCheckResult;
+
+    /// Cheap feasibility check on the current trail — no model, no theory
+    /// combination. Impls may return `Unchanged` to skip on their own policy.
+    fn check_partial_trail(&mut self) -> PartialCheckResult;
```
amarshah1: Not sure if this is the best way to do a partial check?

Issue #52, opened 2026-03-04
Regression tests should also include rejection tests


HuStmpHrrr:
do we have a way to do negative regression tests, like rejecting the input queries?

_Originally posted by @HuStmpHrrr in https://github.com/yaspar-org/Sundance-SMT/issues/43#issuecomment-3999403993_
            
Does that mean we could test for unknown?

PR #152, opened 2026-07-13
Panic for non-linear multiplication

amarshah1:
*Issue #, if available:* this commit addresses #52, but this is not the primary issue for the commit

*Description of changes:*

Long found that this would give an error

```
(set-logic ALL)
(declare-const x Int)
(assert (= (* x x) 4))
(check-sat)
```

This is because the call to the arithmetic solver (non-incremental) would default to 0 when it saw a non-linear multiplication term. I have fixed this by panicking at non-linear multiplication. We could also think about handling these as uninterpreted instead of panicking (see #153)


amarshah1:
Also I think we are not filtering out non-arithmetic equalities, but if you do filter those out, this provides a performance regression. I am not sure why

benjaminfjones:
I'm confused by the claim that (= (* 1 y) y) has a non-linear multiplication term in it. Is there a typo in the formula in the description?

I ran it on all 3 arithmetic backends and internal gives `sat`, z3 gives `sat` and z3incremental gives `unsat`.

```
$ ./target/release/sundance-smt --arithmetic z3incremental linear_mult.smt2
unsat

$ ./target/release/sundance-smt --arithmetic z3 linear_mult.smt2
sat

$ ./target/release/sundance-smt --arithmetic internal linear_mult.smt2
sat
```

benjaminfjones:
> I'm confused by the claim that (= (* 1 y) y) has a non-linear multiplication term in it. Is there a typo in the formula in the description?
> 
> I ran it on all 3 arithmetic backends and internal gives `sat`, z3 gives `sat` and z3incremental gives `unsat`.
> 
> ```
> $ ./target/release/sundance-smt --arithmetic z3incremental linear_mult.smt2
> unsat
> 
> $ ./target/release/sundance-smt --arithmetic z3 linear_mult.smt2
> sat
> 
> $ ./target/release/sundance-smt --arithmetic internal linear_mult.smt2
> sat
> ```

I made an attempt at fixing this disagreement issue in https://github.com/yaspar-org/Sundance-SMT/pull/162 . Let me know what you think.

amarshah1:
Sorry I picked bad examples last time, but something like this we were not erroring on before and actually returning the wrong answer:

```
(set-logic ALL)
(declare-const x Int)
(assert (= (* x x) 4))
(check-sat)
```

benjaminfjones (approved):

PR #162, opened 2026-07-21
Fix short circuit logic in the eager arithmetic paths

benjaminfjones:
*Description of changes:*

The query `(not (= (* 1 y) y))` is unsat, but the eager internal and z3 backends wrongly returned sat. The lone atom is a disequality, which `extract_linear_constraints` intentionally doesn't turn into a LinearConstraint (disequalities are meant to go through Nelson-Oppen). That left constraints and arithmetic_literals both empty, so the guard if constraints.is_empty() && arithmetic_literals.is_empty() short-circuited to `ArithResult::None` before the loop that emits each arithmetic term's definitional equality (`var_{(* 1 y)} == 1·y`). Without that definition the egraph never learned `(* 1 y) = y`, so the disequality was never contradicted. z3incremental was correct because it unconditionally materializes every arithmetic term's definition, so its N-O probe merged the terms and found the conflict.

The solution here tightens the early-return in lialp.rs and z3lp.rs to also require `solver_state.arithmetic_terms.is_empty()`, so the backends register the definitional equalities whenever arithmetic terms exist.

I added `mult_by_one_diseq_unsat.smt2` to the regression suite (passes on all three backends).

----


amarshah1 (approved):
Looks good!

Issue #123, opened 2026-06-18
Update Cadical version if it allows us to not add tautologies


amarshah1:
Currently for each literal `l`, we are giving Cadical the clause `-l l` in order to force Cadical to assign this literal to true/false. We think a newer version of Cadical allows us to force an assignment of a literal without adding this tautology.

We need to confirm if this is the case and if so, update cadical-sys to use the newer version of Cadical

PR #164, opened 2026-07-22
Remove tautology clauses

amarshah1:
*Issue #, if available:* #123

*Description of changes:* An optimization that removes the tautology clauses, i.e. `-l l`. These were originally added because the Cadical version required it to set a specific literal to true, but are no longer required because we updated the Cadical version


amarshah1:
This decreases the total number of clauses learnt (and deleted) which makes sense, but we still learn a lot more clauses than z3 (not sure why)
image
image

amarshah1:
This also enables us to remove the tautology filtering stuff from the proof I captured this in #165 and this can be handled in a seperate PR

src/stats.rs
```diff
@@ -33,6 +33,7 @@ pub struct RoundStats {
     pub egraph_merges: u64,
     pub bool_vars: u64,
     pub clauses: u64,
+    pub binary_clauses: u64,
     pub deleted_clauses: u64,
```
amarshah1, line 37: Addes some more stats. This is not necessary for the remove-tautology-clauses commit, but squeezing it in here

src/egraphs/basic/egraph.rs
```diff
@@ -473,8 +473,7 @@ impl Egraph {
         highest_level: usize,
         highest_hash: u32,
     ) -> (u32, usize, u32) {
-        let p = self.proof_forest[x as usize].clone();
-        match p {
+        match &self.proof_forest[x as usize] {
```
amarshah1, line 476: This is also just a slight optimization/simplification, but not related to removing the tautology at all

HuStmpHrrr:
don't you need to also change cnf too?

amarshah1:
I don't think so? What changes to cnf would we need to make?

benjaminfjones (approved):

HuStmpHrrr:
nvm, maybe I hallucinated.

PR #166, opened 2026-07-25
Implement gate-context Boolean propagation

amarshah1:
*Issue #, if available:*

*Description of changes:* This PR implements a Z3-style gate_ctx optimization for Boolean terms. Boolean terms used only as propositional gates are no longer merged with true or false, reducing unnecessary egraph merges.
To preserve soundness, the egraph tracks SAT-assigned Boolean values on e-class roots and detects conflicts when classes with opposite values are merged.


amarshah1:
This offers a very slight performance improvement, but reduces the number of egraph merges significantly

image

image

benjaminfjones:
Can you explain what gate-context booleans are? Is there a reference you can point me to?

src/egraphs/basic/egraph.rs
```diff
@@ -888,6 +969,136 @@ impl Egraph {
         EgraphResult::ok()
     }
 
+    fn assign_bool_value(
```
benjaminfjones, line 972: top-level comments on the next three functions would be good

src/solver_state.rs
```diff
@@ -13,22 +13,40 @@ use std::collections::{HashMap, HashSet};
...
 use crate::egraphs::basic::egraph::Egraph;
-use crate::egraphs::traits::EgraphTrait;
+use crate::egraphs::traits::{EgraphResult, EgraphTrait};
 use crate::solver_types::{
     Assertion, ConstructorType, ConstructorType::*, Polarity, Quantifier, TermOption,
 };
 
 use crate::utils::DeterministicHashMap;
 
+fn is_gate(term: &Term, context: &mut Context) -> bool {
```
benjaminfjones, line 32: Consider commenting here about what gate means and where/why it's useful

benjaminfjones:
What's the high level motivation for this change? From the cactus plot is doesn't appear to improve performance. I was also wondering about the # of queries solved in the timeout before & after.

Is this stability related, or code complexity/simplicity? Or something else... ?

src/solver_state.rs
```diff
@@ -13,22 +13,40 @@ use std::collections::{HashMap, HashSet};
...
+    match term.repr() {
+        And(_)
+        | Or(_)
+        | Not(_)
+        | Implies(_, _)
+        | Xor(_)
+        | Ite(_, _, _)
```
benjaminfjones, line 39: There's a markdown file in the project root `egraph_term_registration_comparison.md`  (I'm guessing is LLM gen'd) that says Ite is not a gate, but here it's marked as such. What's the case?

Also, I'm guessing all the three children get the gate context as well?

Also, I'd suggest either removing the design doc `egraph_term_registration_comparison.md` or moving it to a docs directory if you think it's worth keeping and it's accurate.

src/egraphs/basic/egraph.rs
benjaminfjones: Nevermind, I see they're documented on the trait.

benjaminfjones:
Approving so you're not blocked. If you can add some comments and context around the gate booleans that would be great. I'm out on VAC until this coming Thursday.

benjaminfjones (approved):

src/cadical_propagator.rs
```diff
@@ -43,6 +42,38 @@ pub struct CustomExternalPropagator<'a> {
...
+
+    fn queue_egraph_conflict(&mut self, conflict: &Conflict<u32>) {
+        let clause = self.solver_state.egraph_conflict_clause(conflict);
+        // A tautological conflict clause is always satisfied, so it rules out
+        // nothing; skip it. CaDiCaL would ignore it as an external clause anyway.
+        if Self::is_tautological(&clause) {
+            return;
+        }
+        self.disequalities.borrow_mut().push(clause);
+    }
```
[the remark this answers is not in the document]
amarshah1: This doesn't really make sense, but we really should not have tautological conflicts. I think the best move might just be to panic if there is one

amarshah1:
>  Can you explain what gate-context booleans are? Is there a reference you can point me to?

Not really, this is just an optimization in Z3's codebase, see [here](https://github.com/Z3Prover/z3/blob/a9115d9de1c0054babca5e9ebee3436c0d0c5f03/src/smt/smt_internalizer.cpp#L31-L48) 

The high-level idea is that certain boolean terms are not needed in congruence closure, because they never occur as the input of a function or inside an equality. Thus, we can treat these terms as "gates" and save congruence closure time

> What's the high level motivation for this change? From the cactus plot is doesn't appear to improve performance. I was also wondering about the # of queries solved in the timeout before & after.

This causes a decrease in the number of merges, which I think can matter a lot for performance. I think right now on the Verus benchmarks we are hitting some sort of bottleneck where the quantifier instantiation order matters so much more than anything else I try

amarshah1:
I'm going to wait to merge this in because the improvements are pretty marginal and I think there are things that can lead to way better performance optimizations

PR #167, opened 2026-07-25
Refactor branch and bound

benjaminfjones:
*Description of changes:*

The two changes together rewrite `src/arithmetic/lia/lira_solver.rs`, replacing
the old slotmap-arena branch-and-bound explorer (~230 lines) with a
straightforward DFS whose control flow mirrors the math: at each node, find the
first fractional int var, explore floor then ceil, and let the parent resolve
its own two children's conflicts.

The first pass (commit) expressed this as function-call recursion; the second
converted it to an explicit heap-allocated stack of BranchFrames driven by a
Stage state machine (ExploreFloor/AwaitFloor/ExploreCeil/AwaitCeil). Search
depth is now bounded by heap memory rather than the OS thread stack. Along the
way the rewrite eliminated per-node Conflict clones (conflicts move up the
stack) and added a pruning improvement: if a child's conflict doesn't mention
the branching variable, the sibling branch is skipped entirely.

The public API (`LIRASolver::new`, `solve()`, `lra_solver()/lra_solver_mut()`) is unchanged.

The re-write improves number of UNSAT queries solved in the full benchmark suite by 2.2% when using
internal arithmetic.

----

### DIFF
--- a/src/arithmetic/lia/config.rs
+++ b/src/arithmetic/lia/config.rs
@@ -13,6 +13,13 @@ pub struct SolverConfig {
     /// Upper bound on the number of LRA solve calls in branch-and-bound.
     /// `None` means unlimited.
     pub max_lra_solve_calls: Option<usize>,
+    /// Maximum branch-and-bound tree depth explored before giving up.
+    ///
+    /// The branch-and-bound search is recursive, so this doubles as a stack-overflow
+    /// guard: a node at depth `d` sits `d` frames deep on the call stack. When the limit
+    /// is reached the search returns `UNKNOWN` rather than recursing further. `None`
+    /// means unlimited (only safe when running on a thread with a large stack).
+    pub max_branch_depth: Option<usize>,
 }
 
 impl Default for SolverConfig {
@@ -24,6 +31,9 @@ impl Default for SolverConfig {
             // significantly, some will return UNKNOWN instead since branch-and-bound
             // will return UNKNOWN when the limit is hit.
             max_lra_solve_calls: None,
+            // Deep enough that well-behaved problems never hit it, low enough to stay
+            // clear of the default thread stack limit on pathological inputs.
+            max_branch_depth: Some(4096),
         }
     }
 }
--- a/src/arithmetic/lia/config.rs
+++ b/src/arithmetic/lia/config.rs
@@ -15,10 +15,10 @@ pub struct SolverConfig {
     pub max_lra_solve_calls: Option<usize>,
     /// Maximum branch-and-bound tree depth explored before giving up.
     ///
-    /// The branch-and-bound search is recursive, so this doubles as a stack-overflow
-    /// guard: a node at depth `d` sits `d` frames deep on the call stack. When the limit
-    /// is reached the search returns `UNKNOWN` rather than recursing further. `None`
-    /// means unlimited (only safe when running on a thread with a large stack).
+    /// Branch-and-bound uses an explicit heap-allocated stack (not recursion), so depth is
+    /// bounded by heap memory rather than the OS thread stack; this limit can therefore be
+    /// set high without risking stack overflow. When it is reached the search returns
+    /// `UNKNOWN` rather than descending further. `None` means unlimited.
     pub max_branch_depth: Option<usize>,
 }
 
@@ -31,9 +31,10 @@ impl Default for SolverConfig {
             // significantly, some will return UNKNOWN instead since branch-and-bound
             // will return UNKNOWN when the limit is hit.
             max_lra_solve_calls: None,
-            // Deep enough that well-behaved problems never hit it, low enough to stay
-            // clear of the default thread stack limit on pathological inputs.
-            max_branch_depth: Some(4096),
+            // Branch-and-bound runs on an explicit heap stack, so this is no longer capped by
+            // the thread stack. Match `max_lra_solve_calls`' historical 2^17 headroom so
+            // regression tests reach SAT/UNSAT or TIMEOUT rather than bailing to UNKNOWN.
+            max_branch_depth: Some(1 << 17),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/arithmetic/lia/config.rs
```diff
@@ -13,6 +13,13 @@ pub struct SolverConfig {
     /// Upper bound on the number of LRA solve calls in branch-and-bound.
     /// `None` means unlimited.
     pub max_lra_solve_calls: Option<usize>,
+    /// Maximum branch-and-bound tree depth explored before giving up.
+    ///
+    /// Branch-and-bound uses an explicit heap-allocated stack (not recursion), so depth is
+    /// bounded by heap memory rather than the OS thread stack; this limit can therefore be
+    /// set high without risking stack overflow. When it is reached the search returns
+    /// `UNKNOWN` rather than descending further. `None` means unlimited.
```
amarshah1, line 21: when possible, can you try to make the comments more concise?

src/arithmetic/lia/lira_solver.rs
```diff
@@ -20,547 +18,502 @@ use crate::arithmetic::lia::stats::Stats;
...
-                    SolverDecision::UNKNOWN,
-                    self.stats.clone(),
-                ));
+    /// Branch-and-bound driven by an explicit heap stack instead of the call stack.
+    ///
+    /// This is a direct iterative rendering of the recursive descent: each `BranchFrame` on
+    /// `stack` is one node whose relaxation was feasible-but-fractional, and `stage` records
+    /// which child it is exploring. Because depth is now bounded by heap memory rather than
+    /// the OS thread stack, [`SolverConfig::max_branch_depth`] can be set far higher than a
+    /// recursive version could safely tolerate.
```
amarshah1, line 192: Does this lead to a performance improvement in practice?

amarshah1:
It might be worth adding tests that check that we give unknown when the max_branch_depth is not high enough?

src/arithmetic/lia/lira_solver.rs
```diff
@@ -20,547 +18,502 @@ use crate::arithmetic::lia::stats::Stats;
...
+    /// explored; the caller must `backtrack(level)` once that child completes.
+    fn step(
+        &mut self,
+        x: Var,
+        side: BranchSide,
+        bound: Integer,
+        depth: usize,
+    ) -> SolverResult<ExploreStep> {
+        // Guard the LRA-solve budget and the (now heap-bounded) branch depth.
+        if self.over_budget(depth) {
```
amarshah1, line 446: since you are checking at the beginning, you are basically saying you can go exactly one less than the budget right?

src/arithmetic/lia/lira_solver.rs
```diff
@@ -20,547 +18,502 @@ use crate::arithmetic::lia::stats::Stats;
...
+                    NodeOutcome::Pruned(ceil_conflict)
+                } else {
+                    // Both branches are infeasible and both depend on `x`. Resolve on `x` to
+                    // obtain a conflict that no longer mentions it, proving this node
+                    // infeasible.
+                    let floor_conflict = stack[top]
+                        .floor_conflict
+                        .take()
+                        .expect("ceil branch completed without a saved floor conflict");
+                    NodeOutcome::Pruned(floor_conflict.resolve(x, &ceil_conflict))
```
amarshah1: Claude is asking about

"When both branches are trivially infeasible on x (fractional relaxation bounds like [2.3, 2.8]), resolve(x, ...) yields an empty conflict, which surfaces as INFEASIBLE(∅) and forces global UNSAT via an empty learned clause in the propagator. This behavior is inherited from the old resolve(), not new here — but it's soundness-critical and untested. Can we confirm an empty core genuinely means unconditional theory-infeasibility, and add a debug_assert!/regression test to pin it?"

I'm not sure if this is a real issue or not

amarshah1 (approved):

src/arithmetic/lia/lira_solver.rs
benjaminfjones: I don't know if the explicit stack approach leads to better performance; I wasn't able to test this in any real benchmarks that have good perf characteristics for benchmarking because at the 4096 (or slightly higher) stack depth, these problems hit the limit and return unknown. If the limit is raised much higher, the stack overflows.

I think the real performance improvement is due to a combination of the early pruning optimization (this alone gave over a percentage point improvement on the # of unsat solved benchmark) and the removal of cloning at each step in conflict resolution.

src/arithmetic/lia/config.rs
benjaminfjones: Yeah, sorry. I tend to write books

src/arithmetic/lia/lira_solver.rs
benjaminfjones: I will have to think about this more carefully. I don't really understand the comment.

benjaminfjones:
Before change verus benchmark run using internal arithmetic:
```
============================================================
SUMMARY
============================================================

Total files processed: 3763
  - UNSAT results:   2794 (74.2%)
  - UNKNOWN results:   18 (0.5%)
  - TIMEOUT results:  950 (25.2%)
  - OTHER results:      1 (0.0%)

  - Neither UNSAT nor UNKNOWN:  951 (25.3%)
```

After the change
```
============================================================
SUMMARY
============================================================

Total files processed: 3763
  - UNSAT results:   2876 (76.4%)
  - UNKNOWN results:   29 (0.8%)
  - TIMEOUT results:  857 (22.8%)
  - OTHER results:      1 (0.0%)

  - Neither UNSAT nor UNKNOWN:  858 (22.8%)
```

There is a 2.2% improvement on UNSAT results

src/arithmetic/lia/lira_solver.rs
benjaminfjones: There's a unit test `branch_and_bound_triangle` that exercises this exact scenario and produces a length-3 conflict, not empty. That's not proof that it's not a subtle soundness bug, but at least makes me feel a bit more confident.

src/arithmetic/lia/lira_solver.rs
benjaminfjones: When visiting the top of the stack, if the `depth >= max` then we stop. So yeah, I think that's the right sign-posting.

src/arithmetic/lia/lira_solver.rs
benjaminfjones: Addressed this and added both a defensive assert and additional tests.

Issue #103, opened 2026-05-20
Incremental arithmetic component


benjaminfjones:
Currently, the internal arithmetic solver is not incremental from the core solver's point of view. It solves problems in a one-shot manner. If it had an efficient incremental interface that allowed adding and removing variables and assertions, Sundance would save time converting problems and building the linear system internally every time it is called.

Additionally, if the arithmetic solver could backtrack over pivots efficiently, it would not be necessary to snapshot (clone) the tableau when performing the unit cube test.

PR #168, opened 2026-07-28
Stage 1/N of an incremental interface to the internal arithmetic solver

benjaminfjones:
*Issue #, if available:* https://github.com/yaspar-org/Sundance-SMT/issues/103

*Description of changes:*

Before this change, the internal arithmetic solver is purely one-shot and is only called once a full model is obtained by the SAT solver.

This change adds the ability to push and pop linear (in)equalities to the internal arithmetic solver. This is accomplished by growing the underlying tableau when constraints are pushed, and relaxing their bounds to +-infinity (unbounded) when popped. This technique is sound and simple to implement, but it means that the tableau grows monotonically so there is a memory and pivot time cost. This can be eliminated later by backtracking over tableau additions if profiling shows regression.

There is a new frontend for the LRA solver which exposes the increment features. It is not yet wired up to anything in the main solver, but there are several unit tests exercising various push/pop patterns. The new frontend can be used to implement an incremental arithmetic solver trait in the future.

----

### DIFF
--- a/src/arithmetic/lia/incremental.rs
+++ b/src/arithmetic/lia/incremental.rs
@@ -31,7 +31,7 @@
 //!
 //! # Scope / limitations of this first frontend
 //!
-//! - **LRA (rational) reasoning only.** Registered variables are integer-sorted, but
+//! - **LRA (Real) reasoning only.** Registered variables are integer-sorted, but
 //!   integrality is *not* enforced: [`IncrementalLraSolver::check`] may report `Sat`
 //!   for a system that has no integer solution. Sound integer reasoning
 //!   (branch-and-bound) is a later increment.
@@ -204,9 +204,11 @@ impl IncrementalLraSolver {
         }
 
         if let Some(def) = definition {
-            // Assert `new_var == def`, i.e. `new_var - def == 0`, as a fresh row. This
-            // is tautological (new_var is fresh), so it is never needed in a conflict
-            // core and is tracked with no literals.
+            // Assert `new_var == def`, i.e. `new_var - def == 0`, as a fresh row tracked with no
+            // SAT literal, since `new_var` is fresh and this equality never constrains previously
+            // registered variables on its own. This is sound only under the convention that UNSAT
+            // cores are reported relative to all active definitions/background equalities, not as a
+            // self-contained refutation over SAT literals alone.
             self.push_relation(
                 ArithConstraint::Eq(
                     ArithExpr::linear(vec![(var_id, IBig::from(1))], IBig::from(0)),
@@ -277,12 +279,8 @@ impl IncrementalLraSolver {
     /// [`Self::mark_model_var`]) appear in the buckets, keyed by their truncated
     /// integer model value. Variables that are unconstrained (never referenced by any
     /// pushed row) default to `0`.
-    pub fn check(&mut self) -> ArithCheckResult {
-        let decision = self
-            .solver
-            .solve()
-            .expect("lra-inc: unexpected solver error")
-            .decision;
+    pub fn check(&mut self) -> SolverResult<ArithCheckResult> {
+        let decision = self.solver.solve()?.decision;
         match decision {
             SolverDecision::FEASIBLE(assignment) => {
                 let mut buckets: DeterministicHashMap<IBig, DeterministicHashSet<VarId>> =
@@ -294,7 +292,7 @@ impl IncrementalLraSolver {
                     buckets.entry(ibig).or_default().insert(*var_id);
                 }
                 debug_println!(21, 0, "[lra-inc] SAT buckets={:?}", buckets);
-                ArithCheckResult::Sat(buckets)
+                Ok(ArithCheckResult::Sat(buckets))
             }
             SolverDecision::INFEASIBLE(conflict) => {
                 let lits: DeterministicHashSet<i32> = conflict
@@ -302,7 +300,7 @@ impl IncrementalLraSolver {
                     .flat_map(|var| self.slack_to_lits.get(var).into_iter().flatten().copied())
                     .collect();
                 debug_println!(21, 0, "[lra-inc] UNSAT core lits={:?}", lits);
-                ArithCheckResult::Unsat(lits.into_iter().collect())
+                Ok(ArithCheckResult::Unsat(lits.into_iter().collect()))
             }
             SolverDecision::UNKNOWN => {
                 // Pure LRA `solve` terminates in FEASIBLE/INFEASIBLE; treat an
@@ -317,7 +315,7 @@ impl IncrementalLraSolver {
                     let ibig = value.to_int().value().clone();
                     buckets.entry(ibig).or_default().insert(*var_id);
                 }
-                ArithCheckResult::Sat(buckets)
+                Ok(ArithCheckResult::Sat(buckets))
             }
         }
     }
@@ -417,7 +415,7 @@ mod tests {
     #[test]
     fn empty_system_is_sat() {
         let mut s = IncrementalLraSolver::new();
-        assert!(is_sat(&s.check()));
+        assert!(is_sat(&s.check().unwrap()));
     }
 
     #[test]
@@ -427,7 +425,7 @@ mod tests {
         // x <= 5
         s.push_constraint(ArithConstraint::Leq(term(x, 1), ArithExpr::constant(5)), 10)
             .unwrap();
-        assert!(is_sat(&s.check()));
+        assert!(is_sat(&s.check().unwrap()));
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


src/arithmetic/lia/sparse.rs
```diff
@@ -131,6 +131,34 @@ impl<V: Zero + Clone + fmt::Debug> Matrix<V> {
...
+    /// solver has not seen before. Insert entries afterwards with [`Self::update_or_insert`].
+    pub fn add_col(&mut self) -> usize {
+        self.cols.push(FxHashMap::default());
+        self.cols.len() - 1
+    }
+
+    /// Iterate over the non-zero `(col, &value)` entries of a row.
+    ///
+    /// Returns an empty iterator if `row` is out of bounds.
+    pub fn row_entries(&self, row: usize) -> impl Iterator<Item = (usize, &V)> {
```
amarshah1: does this ever get used?

src/arithmetic/lia/tableau_sparse.rs
```diff
@@ -51,6 +51,33 @@ impl TableauSparse {
...
+    /// Set a single entry, inserting, updating, or (if `val` is zero) removing it.
+    pub fn set_entry(&mut self, row: usize, col: usize, val: Rational) -> TableauResult<()> {
+        self.matrix
+            .update_or_insert(row, col, val)
+            .map(|_| ())
+            .map_err(|e| TableauError(e.0))
+    }
+
+    /// Iterate over the non-zero `(col, &coefficient)` entries of a row.
+    pub fn row_entries(&self, row: usize) -> impl Iterator<Item = (usize, &Rational)> {
```
amarshah1: does this ever get used?

src/arithmetic/lia/incremental.rs
```diff
@@ -0,0 +1,647 @@
...
+    ///
+    /// On `Sat`, only variables registered with `report_in_model` (or marked via
+    /// [`Self::mark_model_var`]) appear in the buckets, keyed by their truncated
+    /// integer model value. Variables that are unconstrained (never referenced by any
+    /// pushed row) default to `0`.
+    pub fn check(&mut self) -> ArithCheckResult {
+        let decision = self
+            .solver
+            .solve()
+            .expect("lra-inc: unexpected solver error")
```
amarshah1: can this return SolverResult<ArithCheckResult> to avoid panicking

src/arithmetic/lia/lra_solver.rs
```diff
@@ -341,15 +442,30 @@ impl LRASolver {
...
+    /// reproduces the snapshot point exactly on the unchanged-structure case (so it is
+    /// behavior-preserving for the existing branch-and-bound usage), while also remaining correct
+    /// when rows/columns were added since the snapshot was taken (`add_relation`). Variables that
+    /// post-date the snapshot are absent from the map and default to zero; their basic slack rows
+    /// are then made consistent by the recomputation step.
     fn restore_assignment(&mut self) {
         let previous_model = self
             .old_assignment
-            .as_ref()
+            .clone()
```
amarshah1: Is there some way we can avoid cloning here?

amarshah1: Claude said: "The clone appears forced only because calculate_assignment(&self) borrows self immutably while the loop needs &mut self. You can likely avoid the clone with let prev = self.old_assignment.take(); … restore … self.old_assignment = prev;, reading non-basics from prev without cloning."

amarshah1:
This can be a separate commit, but it might be worth it to create arithmetic specific performance tests (with push/pop), so we can get a sense of how much slower we are compared to z3 and when you improve incrementality, we can get a sense of how much this helps. One way to potentially generate these tests would be to run Sundance on Verus benchmarks and add in push pops as arithmetic constraints get pushed/popped

src/arithmetic/lia/lra_solver.rs
```diff
@@ -243,6 +244,106 @@ impl LRASolver {
...
+    /// all backtracks. The slack starts unbounded, so the first `assert_*` on it records the old
+    /// bound (`None`), and a later `backtrack` restores it to unbounded automatically.
+    pub fn add_relation(&mut self, rel: Rel<Rational>, slack: Var) -> SolverResult<()> {
+        if self.var_to_idx.contains_key(&slack) {
+            return Err(SolverError(format!(
+                "add_relation: slack variable {slack:?} already exists"
+            )));
+        }
+
+        // Step 1: register any brand-new problem variables as unbounded non-basic columns.
```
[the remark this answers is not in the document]
amarshah1: I'm wondering about this also

src/arithmetic/lia/incremental.rs
```diff
@@ -0,0 +1,647 @@
...
+
+        // Level 1: add x <= 1, making the system infeasible.
+        s.notify_new_decision_level();
+        s.push_constraint(ArithConstraint::Leq(term(x, 1), ArithExpr::constant(1)), 20)
+            .unwrap();
+        assert!(matches!(s.check(), ArithCheckResult::Unsat(_)));
+
+        // Backtrack to level 0: the x <= 1 bound is relaxed and the system is feasible.
+        s.notify_backtrack(0);
+        assert!(is_sat(&s.check()));
```
amarshah1: Can we also test that the relaxed bound is gone?

src/arithmetic/lia/incremental.rs
```diff
@@ -0,0 +1,647 @@
...
+//!   [`LRASolver::assert_lower`]/[`LRASolver::assert_upper`].
+//! - The tableau only ever grows (one row per pushed constraint). On backtrack the
+//!   popped constraint's slack simply has its bound relaxed to `(−∞, +∞)` by the
+//!   solver's bound trail — the row is never physically removed.
+//! - Decision levels are tracked by capturing an [`LRASolver::set_backtrack`] token
+//!   at each level boundary and replaying it via [`LRASolver::backtrack`].
+//!
+//! # Scope / limitations of this first frontend
+//!
+//! - **LRA (rational) reasoning only.** Registered variables are integer-sorted, but
```
amarshah1: Did you mean to say rational here or Real?

src/arithmetic/lia/incremental.rs
```diff
@@ -0,0 +1,647 @@
...
+//!   integrality is *not* enforced: [`IncrementalLraSolver::check`] may report `Sat`
+//!   for a system that has no integer solution. Sound integer reasoning
+//!   (branch-and-bound) is a later increment.
+//! - **`div`/`mod` are unsupported.** An [`ArithExpr`] carrying `divs`/`mods` is
```
amarshah1, line 38: what makes it hard to support div?

amarshah1 (approved):
Looks good if you can address the comments!

src/arithmetic/lia/lra_solver.rs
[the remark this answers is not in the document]
benjaminfjones: The frontend API makes this impossible because slack variables are only allocated upon `add_relation` and are fresh by construction. The only way this could happen is using the internals, e.g. in a unit test, where you add a relation, then extract a ref to the slack (not possible through the API) and then construct terms from that slack to form a new relation and add it.

The defensive code Copilot is suggesting is fine, but unnecessary IMO. It just entails an extra iteration over every relation added for what amounts to a logic error in the frontend API.

src/arithmetic/lia/incremental.rs
benjaminfjones: Meant to say Real, but in practice everything is rational because we're in the linear fragment. I'll fix the comment.

src/arithmetic/lia/incremental.rs
benjaminfjones: It's not hard, I am just leaving it for the next PR

src/arithmetic/lia/lra_solver.rs
benjaminfjones: Yep, will address

src/arithmetic/lia/incremental.rs
benjaminfjones: done

src/arithmetic/lia/incremental.rs
benjaminfjones: done

src/arithmetic/lia/tableau_sparse.rs
benjaminfjones: removed

src/arithmetic/lia/incremental.rs
```diff
@@ -0,0 +1,647 @@
...
+            // is tautological (new_var is fresh), so it is never needed in a conflict
+            // core and is tracked with no literals.
+            self.push_relation(
+                ArithConstraint::Eq(
+                    ArithExpr::linear(vec![(var_id, IBig::from(1))], IBig::from(0)),
+                    def,
+                ),
+                None,
+            )?;
+        }
```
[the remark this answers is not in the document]
benjaminfjones: addressed

PR #169, opened 2026-07-30
Defer congruence closure until egraph rebuild

amarshah1:
## Summary
- defer signature collision handling and induced congruence merges to EgraphTrait::rebuild
- queue newly registered terms and parents affected by explicit unions, then rebuild to a fixed point
- invalidate and requeue signature state on backtracking and rebuild at solver observation boundaries
- add focused tests for registration, dynamic registration, nested closure, conflicts, and backtracking

## Testing
- cargo fmt --all -- --check
- cargo clippy --no-default-features --all-targets -- -D warnings
- cargo test --no-default-features -- --skip regression_test
- cargo test --release --no-default-features --features local-z3 --test regression_test -- --nocapture

Regression result: 417 correct, 0 incorrect, 6 timeouts. The expected-unsat eq_diamond18 and eq_diamond19 cases return unsat with a longer limit. The expected-unknown quantifier_ite_no_error6 case still times out at 120 seconds.

### DIFF
--- a/src/egraphs/basic/egraph.rs
+++ b/src/egraphs/basic/egraph.rs
@@ -17,11 +17,6 @@ use yaspar_ir::ast::Local;
 /// Key for the signature table: (operator, canonical children).
 type SigKey = (CanonicalOp, Children);
 
-/// Trail entry for undoing sig_table modifications on backtrack.
-/// Stores the actual key used, so undo doesn't depend on UF state.
-/// (level, key, term_id, was_inserted)
-type SigTrailEntry = (usize, SigKey, u32, bool);
-
 /// A Boolean value attached to an e-class and its SAT witness. Constants use zero.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct BoolClassAssignment<T> {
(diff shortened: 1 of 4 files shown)


src/egraphs/basic/egraph.rs
```diff
@@ -2064,21 +2157,31 @@ impl EgraphTrait for Egraph {
 
     fn rebuild(&mut self) -> EgraphResult<Self::TermId> {
```
benjaminfjones, line 2158: It looks like every path that bumps the level (notify_new_decision_level, the N-O probe loop in `cadical_propagator.rs`) runs a full rebuild. That looks correct, but is there a possibility that a future change introduces a point where full rebuild is again necessary but not obvious and is forgotten causing a soundness problem? How could we guard against that?

src/egraphs/basic/egraph.rs
```diff
@@ -17,10 +17,8 @@ use yaspar_ir::ast::Local;
 /// Key for the signature table: (operator, canonical children).
 type SigKey = (CanonicalOp, Children);
 
-/// Trail entry for undoing sig_table modifications on backtrack.
-/// Stores the actual key used, so undo doesn't depend on UF state.
-/// (level, key, term_id, was_inserted)
-type SigTrailEntry = (usize, SigKey, u32, bool);
+/// A complete repair pass is worthwhile only when incremental work is dense.
```
benjaminfjones, line 20: Does dense just mean there's enough incremental work to do? Or is this as opposed to some notion of sparse?

benjaminfjones: Did you try other values here to see if they are better/worse?

tests/egraph_rebuild.rs
```diff
@@ -0,0 +1,192 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use sundance_smt::egraphs::{Egraph, EgraphTrait, Op};
+
+fn term(egraph: &mut Egraph, op: &str, children: &[u32]) -> u32 {
+    egraph.register_term(Op::App(op.to_string()), children, false)
+}
+
+#[test]
+fn registration_congruence_is_deferred_until_rebuild() {
```
benjaminfjones, line 11: Great set of tests!

benjaminfjones (approved):

PR #170, opened 2026-07-30
New feedback functions 

LongQianQL:
*Issue #, if available:*

*Description of changes:* This PR adds new logging functionalities to Sundance by implementing two runtime flags:

1. `--trail-out <file>`: This logs the propositional skeletons encountered during the search.
2. `--partial-proof <file>`: This logs the proof file upon termination, even when Sundance does not answer UNSAT. 

### DIFF
--- a/src/cadical_propagator.rs
+++ b/src/cadical_propagator.rs
@@ -40,13 +40,53 @@ pub struct CustomExternalPropagator<'a> {
     /// Incremental Z3 arithmetic state — Some iff `arithmetic == ArithSolver::Z3Incremental`.
     #[cfg(feature = "z3-solver")]
     pub z3_incremental: Option<Z3IncrementalState>,
-    // --trail-out logging (inert unless trail_out_active): |lit| -> atom, and each refuted model
-    pub trail_out_active: bool,
+    // --trail-out logging (inert unless the writer is Some). Trails stream to
+    // disk as they are refuted; the small |lit| -> atom map is held and flushed
+    // at the end (only complete then, as new literals appear during the search).
+    pub trail_writer: Option<std::io::BufWriter<std::fs::File>>,
     pub trail_atoms: std::collections::HashMap<i32, String>,
-    pub refuted_trails: Vec<Vec<i32>>,
 }
 
 impl<'a> CustomExternalPropagator<'a> {
+    /// Stream one refuted model as a `t <signed lits>` line. A write error is
+    /// reported once then the writer is dropped; it must never abort the solve.
+    fn write_trail_line(&mut self, model: &[i32]) {
+        use std::io::Write;
+        let Some(w) = self.trail_writer.as_mut() else {
+            return;
+        };
+        let mut line = String::with_capacity(model.len() * 5 + 2);
+        line.push('t');
+        for lit in model {
+            use std::fmt::Write as _;
+            let _ = write!(line, " {lit}");
+        }
+        if let Err(e) = writeln!(w, "{line}") {
+            debug_println!(2, 0, "Failed to stream trail line: {}", e);
+            self.trail_writer = None; // stop trying after the first failure
+        }
+    }
+
+    /// Append the `m <id> <atom>` map (sorted by id) and close the trail log.
+    /// Called once after the solve, when the literal set is finally complete.
+    pub fn finish_trail_log(&mut self) {
+        use std::io::Write;
+        let Some(mut w) = self.trail_writer.take() else {
+            return;
+        };
+        let mut ids: Vec<&i32> = self.trail_atoms.keys().collect();
+        ids.sort();
+        let res = (|| -> std::io::Result<()> {
+            for id in ids {
+                writeln!(w, "m {} {}", id, self.trail_atoms[id])?;
+            }
+            w.flush()
+        })();
+        if let Err(e) = res {
+            debug_println!(2, 0, "Failed to write trail atom map: {}", e);
+        }
+    }
+
     /// Register any new CNF variables created since the last sync.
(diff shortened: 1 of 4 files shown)


src/cdcl.rs
```diff
@@ -158,6 +163,63 @@ pub fn cdcl_decision_procedure(
         }
     }
 
+    // Dump the proof forest for any result: complete on unsat, else a prefix (no empty clause)
+    if let Some(p) = partial_proof_file {
+        let complete = result == Status::UNSATISFIABLE;
```
amarshah1: can we factor the inside of the conditional into a helper function

src/cdcl.rs
```diff
@@ -158,6 +163,63 @@ pub fn cdcl_decision_procedure(
...
+                "{} proof forest written to: {}",
+                if complete { "Complete" } else { "Partial" },
+                p.display()
+            );
+        }
+    }
+
+    // Write the refuted-model trail: a `<id> <atom>` map, a blank line, then `<signed lits> 0` per model
+    if let Some(p) = trail_file {
+        let mut out = String::new();
```
amarshah1: could this also be factored into a helper function

src/cadical_propagator.rs
```diff
@@ -368,6 +372,18 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
     }
 
     fn cb_check_found_model(&mut self, model: &[i32]) -> bool {
+        // --trail-out: every model seen here in a non-SAT run is refuted, so record it at entry
+        if self.trail_out_active {
+            for &l in model {
+                let id = l.unsigned_abs() as i32;
+                if !self.trail_atoms.contains_key(&id) {
```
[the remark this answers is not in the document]
LongQianQL: This is intended as the trail is a log of visited propositional skeletons found by the SAT solver.

src/cadical_propagator.rs
```diff
@@ -40,6 +40,10 @@ pub struct CustomExternalPropagator<'a> {
     /// Incremental Z3 arithmetic state — Some iff `arithmetic == ArithSolver::Z3Incremental`.
     #[cfg(feature = "z3-solver")]
     pub z3_incremental: Option<Z3IncrementalState>,
+    // --trail-out logging (inert unless trail_out_active): |lit| -> atom, and each refuted model
+    pub trail_out_active: bool,
+    pub trail_atoms: std::collections::HashMap<i32, String>,
+    pub refuted_trails: Vec<Vec<i32>>,
```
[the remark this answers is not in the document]
LongQianQL: `trail_atoms` should not be a problem since the atoms are tracked internally anyways I think. The logging of `refuted_trails` can get large and expensive over long runs, but the target use case of this is a partial `10-20s` run, so this should be fine.

src/cdcl.rs
```diff
@@ -158,6 +163,63 @@ pub fn cdcl_decision_procedure(
         }
     }
 
+    // Dump the proof forest for any result: complete on unsat, else a prefix (no empty clause)
+    if let Some(p) = partial_proof_file {
+        let complete = result == Status::UNSATISFIABLE;
+        let status = match result {
```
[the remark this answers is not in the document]
LongQianQL: I didn't consider the possibility of `--proof --partial-proof` being called concurrently, maybe this should just be disabled?

src/cdcl.rs
```diff
@@ -158,6 +163,63 @@ pub fn cdcl_decision_procedure(
...
+            Status::UNSATISFIABLE => "unsat",
+            Status::SATISFIABLE => "sat",
+            Status::UNKNOWN => "unknown",
```
amarshah1: I think you will never get unknown at this point because this is the result that Cadical gives. Sundance converts sat to unknown in main.rs

src/cadical_propagator.rs
```diff
@@ -368,6 +372,18 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
...
+        // --trail-out: every model seen here in a non-SAT run is refuted, so record it at entry
+        if self.trail_out_active {
+            for &l in model {
+                let id = l.unsigned_abs() as i32;
+                if !self.trail_atoms.contains_key(&id) {
+                    let atom = format!("{}", self.solver_state.get_term_from_lit(id));
+                    self.trail_atoms.insert(id, atom);
+                }
+            }
+            self.refuted_trails.push(model.to_vec());
```
amarshah1: Could the models can be streamed into a file as the solver runs instead of pushed into refuted_trails? I'm worried we might get an OOM on queries that run a while

amarshah1:
For some reason the linter doesnt like when you use `<>` in comments

LongQianQL:
Ah I see, I think this should be fixed now.

amarshah1 (approved):
I think this is ok, can you make sure to address all the comments. If someone enable both proof and partial proof, maybe we can either print a warning or panic?

src/cdcl.rs
LongQianQL: I think this can happen if `--timeout` is used or the solver is interrupted. I've updated the header to reflect this.

src/cadical_propagator.rs
LongQianQL: This sounds good, I've implemented this.

LongQianQL:
Thanks! I've updated the branch to address the comments, the solver now panics when `--proof --partial-proof` are both used.

src/cdcl.rs
LongQianQL: Thanks, I will do this.

src/cdcl.rs
LongQianQL: Thanks! I will implement this.

PR #172, opened 2026-08-03
Batching quantifier instantiations

amarshah1:
*Issue #, if available:*

*Description of changes:* Adding an option to bundle quantifier instantiations. Default is 85 since that performs well empirically

### DIFF
--- a/src/cadical_propagator.rs
+++ b/src/cadical_propagator.rs
@@ -37,6 +37,8 @@ pub struct CustomExternalPropagator<'a> {
     /// Once reached, stop probing further pairs even if unprobed pairs remain.
     pub max_arith_conflicts_per_round: usize,
     pub last_observed_var: i32,
+    /// Max instantiations to materialize per complete-model check. 0 = unbounded.
+    pub batch_cap: usize,
     /// Incremental Z3 arithmetic state — Some iff `arithmetic == ArithSolver::Z3Incremental`.
     #[cfg(feature = "z3-solver")]
     pub z3_incremental: Option<Z3IncrementalState>,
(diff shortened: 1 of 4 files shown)


src/quantifiers/quantifier.rs
```diff
@@ -134,6 +134,25 @@ pub(crate) fn instantiate_quantifiers(
...
+                        .map(|(k, v)| (k.clone(), solver_state.get_term(solver_state.to_solver_uid(*v))))
+                        .collect();
+                    let dup = solver_state
+                        .added_instantiations
+                        .get(&quantifier.id)
+                        .map(|s| s.contains(&subs))
+                        .unwrap_or(false);
+                    if !dup { newc += 1; }
+                }
+                eprintln!("MATCH qid={} total={} new={}", quantifier.id, total, newc);
```
HuStmpHrrr: is it expected to use eprintln instead of a logging output?

src/quantifiers/quantifier.rs
```diff
@@ -134,6 +134,25 @@ pub(crate) fn instantiate_quantifiers(
                 continue;
             }
 
+            // INSTRUMENTATION: count total vs new (non-dup) matches per quantifier
+            if std::env::var("SUNDANCE_MATCH_STATS").is_ok() {
```
HuStmpHrrr: do we want to repeatedly query env var? same question for all env vars below.

HuStmpHrrr (approved):

src/quantifiers/quantifier.rs
amarshah1: Yeah sorry I added this for some debugging stuff, but I just undid that commit

src/quantifiers/quantifier.rs
amarshah1: Same as other comment, I undid the commit with teh debugging stuff

Issue #176, opened 2026-08-07
Soundness Bug - Inductive Datatypes


LongQianQL:
Sundance incorrectly returns `UNSAT` for the following formula, which should be `SAT` with `b = false, x = Nil, tail(Nil) = Nil`.

```
(set-logic ALL)
(declare-datatypes ((IntList 0)) (((Nil) (Cons (head Int) (tail IntList)))))
(declare-const x IntList)
(declare-const b Bool)
(assert (=> b ((_ is Cons) x)))
(assert (= (tail x) x))
(check-sat)
```

PR #181, opened 2026-08-08
Guard datatype occurs-check conflict clause with constructor testers

amarshah1:
*Issue #, if available:* #176 

*Description of changes:* The occurs check for inductive datatypes detects a cycle in the constructor-child graph and learns a conflict clause negating the equalities that link the cycle. But a node only participates in the cycle because its tester (is-Ctor) holds, and that dependency was dropped from the clause.

For a formula like (=> b (is-Cons x)) with (= (tail x) x), the solver decides is-Cons(x)=true, builds the edge x -> (tail x), closes the cycle via (tail x) = x, and then learned the unconditional clause (tail x) != x. That wrongly rules out the model b=false, x=Nil, so a SAT formula was reported unsat.

Carry each parent's tester literal along its graph edge and add the negated tester of every constructor on the cycle to the conflict clause. The clause becomes e.g. (or (not (= (tail x) x)) (not (is-Cons x))), which is sound and preserves the x=Nil model.

Add regression tests (sat and unsat) covering guarded self-loops, multi-node guarded cycles, tree and polymorphic variants, and the forced-cycle unsat companions.


src/datatypes/occurs_check.rs
```diff
@@ -11,10 +11,12 @@ use crate::solver_types::ConstructorType;
...
-/// An edge in the constructor graph: (parent_eid, child_eid).
-/// parent_eid is the egraph ID of the constructor term,
-/// child_eid is the egraph ID of its recursive child (selector application).
-type Edge = (u32, u32);
+/// An edge in the constructor graph: (parent_eid, child_eid, parent_tester_lit).
+/// parent_eid/child_eid are the egraph IDs of the constructor term and its
+/// recursive child. parent_tester_lit is the `(_ is Ctor) parent` literal that
+/// makes the parent this constructor; the conflict clause must be guarded by its
+/// negation. `None` if no tester literal exists yet (nothing to guard).
+type Edge = (u32, u32, Option<i32>);
```
benjaminfjones: Consider using a struct here. Once we get past 2-tuples it gets hard to read/reason about. Not that we're reading code anymore, haha

benjaminfjones (approved):

Issue #177, opened 2026-08-07
Soundness bug - Real sorts


LongQianQL:
Sundance incorrectly returns `SAT` for the following formula. It appears that terms of sort `Real` are not registered as arithmetic terms. 

```
(set-logic ALL)
(declare-const a Real)
(declare-const b Real)
(assert (not (= a b)))
(assert (<= a b))
(assert (<= b a))
(check-sat)
```

PR #182, opened 2026-08-10
Panic on Real arithmetic instead of returning unsound results

amarshah1:
*Issue #, if available:* #177

*Description of changes:* Real-sorted terms are never registered with the arithmetic solver. `register_arithmetic_and_quantifier` only checks for the `Int` sort, so `Real` terms fall through and are treated as opaque uninterpreted constants.

This PR, adds a stopgap guard in `register_arithmetic_and_quantifier` that panics on Real-sorted terms.

### DIFF
--- a/src/solver_state.rs
+++ b/src/solver_state.rs
@@ -509,7 +509,20 @@ impl SolverState {
         let num = term.uid();
 
         // Arithmetic term tracking
-        if term.get_sort(self.context.arena()).to_string() == "Int" {
+        let sort = term.get_sort(self.context.arena()).to_string();
+        // TODO: Real arithmetic is not yet plumbed through the arithmetic
+        // backends (they allocate integer variables and bucket Nelson-Oppen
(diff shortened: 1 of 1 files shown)


LongQianQL (approved):

src/solver_state.rs
```diff
@@ -509,7 +509,20 @@ impl SolverState {
         let num = term.uid();
 
         // Arithmetic term tracking
-        if term.get_sort(self.context.arena()).to_string() == "Int" {
+        let sort = term.get_sort(self.context.arena()).to_string();
+        // TODO: Real arithmetic is not yet plumbed through the arithmetic
```
benjaminfjones, line 513: Can we make this into a tracked issue?

benjaminfjones (approved):

src/solver_state.rs
amarshah1: Yeah it is #183

Issue #112, opened 2026-05-27
Architectural concerns regarding eDRAT proof production


amazoniancodel:
This issue addresses how and where Sundance makes eDRAT proofs. My proposed changes to the eDRAT format will require standardizing how Sundance logs theory lemmas with the proof tracker. The discussion for how this will affect Sundance will live here.

## The current state of the world

Currently, Sundance adds clauses to the accumulated eDRAT proof in two ways. Most proof steps are added by CaDiCaL when it invokes a callback mechanism in `proof_tracer.rs`. For example, the [`add_original_clause()`](https://github.com/yaspar-org/Sundance-SMT/blob/e18a2e92758c42cfbdc06e0bc89dbf747bbaf561/src/proof/proof_tracer.rs#L325) function adds to the proof original SMT clauses and theory lemmas. The two are differentiated by a boolean flag `finished_original_clauses`, which is toggled in `cdcl.rs` [here](https://github.com/yaspar-org/Sundance-SMT/blob/e18a2e92758c42cfbdc06e0bc89dbf747bbaf561/src/cdcl.rs#L95). SAT learned clauses are also added to the eDRAT proof this way.

Sundance adds the remaining proof steps manually during the CEGAR loop with the SAT solver. The main place where this is done is during Skolemization, where the Skolem clause is directly added to the proof during instantiation. See [here](https://github.com/yaspar-org/Sundance-SMT/blob/e18a2e92758c42cfbdc06e0bc89dbf747bbaf561/src/quantifiers/quantifier.rs#L141). Along the way, the instantiation lemmas are registered with CaDiCaL, which causes the callback to be invoked and the lemmas to be added to the proof.

## Proposed changes to eDRAT

To support mixed theories, Skolemization, and quantifier instantiation in eDRAT, I propose updating the eDRAT format to include theory IDs in theory lemmas and explicit rules for Skolemization and quantifier instantiation.
```
t <theory_id> <clause> 0
(declare-skolem <id> <name> <term>)
```
However, if we are to accomplish this in Sundance, we need to know: (i) which theory generated the theory lemma, and (ii) which term the skolem variable belongs to. This cannot currently be accomplished with Sundance, since the theory lemmas are added by CaDiCaL, after any theory information has been discarded.

## Proposed changes to Sundance

We should standardize how clauses are registered with CaDiCaL and the eDRAT proof. We should use one set of API functions, rather than manually adding the clauses to both the proof and the CNF formula separately.

[...]

PR #184, opened 2026-08-10
Fixes holes in eDRAT proof production, so that we can support proof checking with Valido + external solver

amarshah1:
*Issue #, if available:* #112

*Description of changes:* This fixes gaps in eDRAT proof production so Valido can check Sundance proofs with an external theory solver.

1. Records otherwise-untracked CaDiCaL original-clause and constraint callbacks as background theory clauses. Theory-specific provenance remains follow-up #186.
2. Tracks expected original-clause callbacks by normalized clause to avoid duplicate or misclassified proof steps.
3. Generates complete SMT-LIB preludes, including dependency-ordered `define-fun` commands and quoted symbols.
4. Preserves required sort ascriptions for polymorphic datatype constructors.
5. Emits each Skolem parent literal before `declare-skolem`, and the child literal afterward.
6. Adds focused regression tests for callback classification, SMT-LIB formatting, dependency ordering, and Skolem ordering.

### DIFF
--- a/src/cadical_propagator.rs
+++ b/src/cadical_propagator.rs
@@ -178,8 +178,23 @@ impl<'a> CustomExternalPropagator<'a> {
             let eq_lit = self.solver_state.get_or_allocate_lit_for_term(&eq_term);
             let clause = vec![lt_lit, gt_lit, eq_lit];
             self.sync_new_vars();
-            self.disequalities.borrow_mut().push(clause);
+            self.queue_theory_clause(clause, Theory::QfLia);
         }
+    }
+
+    /// Queues a clause whose proof step has already been recorded.
+    fn queue_external_clause(&self, clause: Vec<i32>) {
+        self.proof_tracer
+            .borrow_mut()
+            .register_clause_for_cadical_callback(&clause);
+        self.disequalities.borrow_mut().push(clause);
+    }
+
+    fn queue_theory_clause(&self, clause: Vec<i32>, theory: Theory) {
+        self.proof_tracer
+            .borrow_mut()
+            .add_theory_clause(&clause, theory);
+        self.queue_external_clause(clause);
     }
 
     pub fn sync_external_stats(&mut self) {
@@ -204,10 +219,7 @@ impl<'a> CustomExternalPropagator<'a> {
                 Skolemization { clauses } => clauses,
             };
             for clause in clauses {
-                self.proof_tracer
-                    .borrow_mut()
-                    .expect_original_clause_callback(clause);
-                self.disequalities.borrow_mut().push(clause.clone());
+                self.queue_external_clause(clause.clone());
             }
         }
         // Materializing an instance can enqueue arithmetic merges (via
@@ -332,15 +344,8 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
                         self.disequalities.borrow()
                     );
 
-                    // CC TODO: Are these congruence closure/EUF atoms? Comment just below this one suggests so.
-                    {
-                        let mut proof_tracer = self.proof_tracer.borrow_mut();
-                        proof_tracer.add_theory_clause(&shrunk_constraint, Theory::Background);
-                        proof_tracer.expect_original_clause_callback(&shrunk_constraint);
-                    }
-
                     // let theory_reason = format!("congruence_closure_level_{}", self.decision_level);
-                    self.disequalities.borrow_mut().push(shrunk_constraint);
+                    self.queue_theory_clause(shrunk_constraint, Theory::Background);
                     debug_println!(
                         14 - 3,
                         0,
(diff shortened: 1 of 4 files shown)


amarshah1:
PR #181 has merged, and this PR now targets `main`.

src/proof/proof_tracer.rs
```diff
@@ -53,6 +54,13 @@ where
     false
 }
 
+fn normalized_clause(clause: &[i32]) -> Vec<i32> {
```
amazoniancodel: Nitpick - verb-first naming convention says that this should be `normalize_clause()` or `get_normalized_clause()`, but this is not blocking/threatening correctness.

src/proof/proof_callbacks.rs
```diff
@@ -66,8 +62,9 @@ impl ProofTracer for SMTProofTracer {
         panic!("Do not currently support assumptions")
     }
 
-    fn add_constraint(&mut self, _clause: &[i32]) {
-        // Optional: Adding constraints
+    fn add_constraint(&mut self, clause: &[i32]) {
+        // TODO: Preserve theory provenance for CaDiCaL constraints.
+        self.add_theory_clause(clause, Theory::Background);
```
amazoniancodel: Should the theory be `Background` here? Basically, I wonder when `add_constraint()` gets called. Of course, it's called as a callback by the owned CaDiCaL process, but is the callback mainly driven by CaDiCaL (i.e., it derives the clause), or by Sundance?

src/proof/proof_tracer.rs
```diff
@@ -150,43 +230,64 @@ impl SMTProofTracer {
...
+    pub fn add_theory_clause(&mut self, clause: &[i32], theory: Theory) {
         self.push_step(clause, ProofStepType::TheoryClause(theory));
     }
 
+    pub fn expect_original_clause_callback(&mut self, clause: &[i32]) {
+        let clause = normalized_clause(clause);
+        *self.expected_original_clauses.entry(clause).or_default() += 1;
+    }
+
+    pub fn consume_expected_original_clause(&mut self, clause: &[i32]) -> bool {
```
amazoniancodel: Another nitpick - since `cdcl.rs` line 239 calls `expect_original_clause_callback()` for *every* clause, and not just "original CNF clauses," the name is off. Perhaps this should be `register_clause_for_cdcl_callback()` or something like that.

src/proof/proof_tracer.rs
```diff
@@ -150,43 +230,64 @@ impl SMTProofTracer {
...
+        let clause = normalized_clause(clause);
+        match self.expected_original_clauses.get_mut(&clause) {
```
amazoniancodel: This `match` block can be simplified slightly by combining the `Some` branches. If the function's return value is `Some`, check if the original value was 1. If so, call `self.expected_original_clauses.remove(&clause);`, then decrement, i.e.:
```
Some(count) => {
  if *count == 1 {
    self.expected_original_clauses.remove(&clause);
  }
  *count -= 1;
  true
}
```

Alternatively, the `match` block could be ridded of entirely:
```
if let Some(count) = self.expected_original_clauses.get_mut(&clause) {
  if *count == 1 {
    self.expected_original_clauses.remove(&clause);
  }
  *count -= 1;
  true
} else {
  false
}
```

src/proof/proof_tracer.rs
```diff
@@ -250,7 +349,6 @@ impl SMTProofTracer {
                 );
             }
         }
-        out.push_str(&temp_output);
```
amazoniancodel, line 253: Good catch! This is simpler.

src/cadical_propagator.rs
```diff
@@ -333,9 +333,11 @@ impl<'a> ExternalPropagator for CustomExternalPropagator<'a> {
                     );
 
                     // CC TODO: Are these congruence closure/EUF atoms? Comment just below this one suggests so.
-                    self.proof_tracer
-                        .borrow_mut()
-                        .add_theory_clause(&shrunk_constraint, Theory::Background);
+                    {
```
amazoniancodel: Any reason why this is in its own context block (`{ }`)?

src/cdcl.rs
```diff
@@ -226,28 +221,22 @@ fn write_partial_proof(path: &std::path::Path, result: Status, edrat_proof: &str
...
-    } else {
-        proof_tracer.borrow_mut().add_original_clause(clause);
+    {
+        let mut proof_tracer = proof_tracer.borrow_mut();
+        if let Some(theory) = theory {
+            proof_tracer.add_theory_clause(clause, theory);
+        } else {
+            proof_tracer.add_original_clause(clause);
+        }
+        proof_tracer.expect_original_clause_callback(clause);
```
amazoniancodel: See other comment about the naming here: we call `expect_original_clause_callback()` for all clauses in this function, whether theory or not. Rename to indicate that this is clause registration, which then gets passed to CaDiCaL shortly after?

amazoniancodel (approved):
Overall looks good! I left several comments with nitpicks about code style and naming convention, but it looks correct, modulo specifying which theories the theory lemmas correspond to.

src/proof/proof_tracer.rs
amarshah1: ok done!

src/proof/proof_tracer.rs
amarshah1: yeah thats way better!

src/proof/proof_callbacks.rs
amarshah1: Yeah it should not be called at all. I added a panic

PR #185, opened 2026-08-10
Clamp non basics pivoted during backtrack

benjaminfjones:
*Description of changes:*

Bug fix in the core LRA solver that impacts re-use of the inner solver during incremental arithmetic.

`restore_assignment` wrote snapshot values into non-basic columns
verbatim. A variable that was *basic* at snapshot time, where being out
of bounds is normal, may since have been pivoted to *non-basic*.
Restoring its stale basic value then puts an out-of-bounds value in a
non-basic column, and `step_simplex` only scans basic variables for
violations, so nothing ever repairs it. `solve` would go on to report
FEASIBLE on a point violating a bound it still had recorded.

Reproducing test: assert `x == 4`, take a backtrack point, solve (which
pivots the `x == 4` slack out of the basis), then backtrack and solve
again -- the model came back `x = 0` with `is_valid()` false, while
`get_bounds` still reported `[4, 4]`.

Fix: clamp each restored non-basic value into its bounds. This loses nothing:
a non-basic variable may sit anywhere within its bounds, and the basics are
recomputed from it immediately afterwards. It is a no-op for variables that were
already non-basic at snapshot time: `backtrack` restores bounds *before* calling
`restore_assignment`, so such a value is checked against the same bounds it was
in-bounds for when snapshotted. Branch-and-bound behavior is therefore unchanged.

Adds a new test: `backtrack_clamps_restored_nonbasic_into_bounds`, which
isolates the basic -> non-basic pivot case; it fails without the clamp.

----

### DIFF
--- a/src/arithmetic/lia/lra_solver.rs
+++ b/src/arithmetic/lia/lra_solver.rs
@@ -454,14 +454,15 @@ impl LRASolver {
 
     /// Restore a previous assignment when backtracking.
     ///
-    /// Restores each **non-basic** variable's value from the snapshot, then recomputes every
+    /// Restores each **non-basic** variable's value from a snapshot, then recomputes every
     /// **basic** variable's value from the current tableau structure so the row equations hold by
-    /// construction. This is pivot-invariant: recomputing basics from the restored non-basics
-    /// reproduces the snapshot point exactly on the unchanged-structure case (so it is
-    /// behavior-preserving for the existing branch-and-bound usage), while also remaining correct
-    /// when rows/columns were added since the snapshot was taken (`add_relation`). Variables that
-    /// post-date the snapshot are absent from the map and default to zero; their basic slack rows
-    /// are then made consistent by the recomputation step.
+    /// construction.
+    ///
+    /// [`Self::solve`] relies on the invariant that every *non-basic* variable is within its
+    /// bounds. A snapshot value is only known to respect that invariant for variables that were
+    /// non-basic *at snapshot time*: a variable that was **basic** then may since have been pivoted
+    /// to non-basic. Restoring its stale basic value verbatim would seat an out-of-bounds value in
+    /// a non-basic column.
     fn restore_assignment(&mut self) {
         // Take the snapshot out of `self` so the loops below can mutably borrow `self` without
         // cloning the map; it is put back at the end so repeated backtracks to the same level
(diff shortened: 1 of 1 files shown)


benjaminfjones:
https://github.com/yaspar-org/Sundance-SMT/commit/9481c33a01b2694e15f297d3f8db75b89a0082fb fixes a soundness bug in how the clamping interacts with the unit cube test. That test contains code that restores the tableau snapshot independent of the backtracking mechanism.

src/arithmetic/lia/lra_solver.rs
```diff
@@ -454,14 +454,33 @@ impl LRASolver {
...
+    /// verbatim would seat an out-of-bounds value in a non-basic column, where nothing would ever
+    /// repair it, and `solve` would report `Feasible` on a point that violates the very bound
+    /// still recorded for that variable.
+    ///
+    /// Clamping to the nearest bound restores the invariant and loses nothing: a non-basic
+    /// variable may sit at any value within its bounds, and the basics are recomputed from it
+    /// immediately below. For a variable that was already non-basic at snapshot time the clamp is
+    /// a no-op — [`Self::backtrack`] restores bounds *before* calling this, so the value is being
+    /// checked against the same bounds it was in-bounds for when snapshotted — which is what keeps
+    /// this behavior-preserving for branch-and-bound.
```
amarshah1: Do you know if it would be possible to make some of these comments a little more concise? I had to read it a few times to understand what was going on

src/arithmetic/lia/lra_solver.rs
```diff
@@ -454,14 +454,33 @@ impl LRASolver {
...
     ///
-    /// Restores each **non-basic** variable's value from the snapshot, then recomputes every
-    /// **basic** variable's value from the current tableau structure so the row equations hold by
-    /// construction. This is pivot-invariant: recomputing basics from the restored non-basics
-    /// reproduces the snapshot point exactly on the unchanged-structure case (so it is
-    /// behavior-preserving for the existing branch-and-bound usage), while also remaining correct
-    /// when rows/columns were added since the snapshot was taken (`add_relation`). Variables that
-    /// post-date the snapshot are absent from the map and default to zero; their basic slack rows
-    /// are then made consistent by the recomputation step.
+    /// Restores each **non-basic** variable's value from the snapshot (clamped into its bounds,
```
amarshah1: My understanding is that in simplex non-basic variables are those that are free to be whatever you choose for them and basic variables are constrained by the values of other variables.

So for example if `s = x + y` is your only constraint, then `x` and `y` are non-basic and s is basic?

amarshah1: But pivoting can swap which variables are basic and non-basic, so the problem you are dealing with is that when you backtrack you need to restore the old non-basic variables back to their bounds and then recheck all of the basic variables?

amarshah1 (approved):
Looks good!

src/arithmetic/lia/lra_solver.rs
benjaminfjones: Yeah, I'll do some more editing.

src/arithmetic/lia/lra_solver.rs
benjaminfjones: > non-basic variables are those that are free to be whatever you choose for them

Yes, but within whatever their current bounds are. It's an invariant of the system that non-basic variables are always in-bounds and basic variables are defined by them (and the tableau coeff's) but are allowed to be outside their bounds.

On the second point, yes, that's right.

src/arithmetic/lia/lra_solver.rs
benjaminfjones: done!

PR #188, opened 2026-08-10
[draft] Fix egraph-ID panic on Boolean bridge literals

amarshah1:
Fixes a `to_egraph_id: no egraph ID for solver uid` panic hit on the Dafny (d_komodo) benchmarks. Targets `pattern-inference`.

## Root cause

When a quantifier is instantiated or skolemized, the instance term is registered in the egraph in its **NNF form** (`insert_predecessor(&nnf_term)`), but a SAT literal is *also* allocated for the **raw, pre-NNF term** as a Boolean bridge (`get_or_allocate_lit_for_term`, in `quantifier.rs`) whenever `t.uid() != nnf_term.uid()`.

NNF rewriting changes the uid — e.g. `(=> true true)` becomes `(or (not true) true)` — so the raw term ends up in `var_map_reverse` (it has a SAT variable) but has **no `id_map` entry** (it was never registered in the egraph). When that bridge literal is later assigned, `process_assignment` looked up the term via `var_map_reverse` and called `to_egraph_id`, which panicked.

## Fix

- Add `to_egraph_id_opt`, returning `Option<u32>` instead of panicking.
- In `process_assignment`, guard both egraph-union sites with it: a literal whose term has no egraph ID is a pure Boolean bridge and carries no theory content (its NNF form does), so the union is safely skipped.

## Verification

- The originally-panicking d_komodo file now passes `:324` (it reaches a separate, pre-existing `lialp.rs:212` div/mod-with-non-constant-divisor limitation later — unrelated).
- 24-file d_komodo batch: **0 `:324` panics** (previously widespread); 1 solves `unsat`, 15 advance to the next blocker, none regress.
- Full regression suite passes (0 failures) — important since the fix is in the hot `process_assignment` path.

Draft pending review of the trigger-inference base (`pattern-inference`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ bundled-z3 = ["z3-solver", "z3/bundled"]
 local-z3 = ["z3-solver"]
 
 [dependencies]
-yaspar-ir = "=2.7.5"
+yaspar-ir = { version = "=2.7.6", features = ["no-pattern"] }
 sat-interface = "0.1.1"
 # Pin the elevate fork until upstream cadical-sys exposes CaDiCaL's elevate option.
 cadical-sys = { git = "https://github.com/amarshah1/cadical-sys.git", rev = "2f234ec95dd36ca74057691cf2fd4e8766a3c40f" }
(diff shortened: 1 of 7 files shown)


amarshah1:
This fixes a bug on the Dafny benchmarks, but need to do more profiling to see if there are any regressions before I merge this in

PR #190, opened 2026-08-13
Make LIRASolver safe to reuse across solve() calls

benjaminfjones:
*Description of changes:*

There are currently two defects that only affect when a single `LIRASolver` serves more than one `solve()` call. The two bugs affect only an incremental frontend.

1. Branch-and-bound leaked speculative branch bounds.

`step` scopes each branch with `set_backtrack()` / ... / `backtrack(level)`. But `LRASolver::set_backtrack` is a no-op while the solver is `Unsat`, and pruning the floor branch during branch-and-bound leaves the system in exactly that state, so the ceil branch's `set_backtrack` returned the current level unchanged, its matching `backtrack` became a no-op too, and the ceil bound survived the search.

2. Stats accumulated across `solve()` calls.

A long-lived solver would eventually exceed any finite budget and report UNKNOWN on every later call regardless of the problem's difficulty. Stats are now reset at `solve()` entry, which also makes the returned `SolverReturn` describe that one solve. The default config sets `max_lra_solve_calls: None`, so this was a latent bug.

----

### DIFF
--- a/src/arithmetic/lia/lira_solver.rs
+++ b/src/arithmetic/lia/lira_solver.rs
@@ -112,7 +112,10 @@ pub struct LIRASolver {
     lra_solver: LRASolver,
     /// Solver configuration.
     config: SolverConfig,
-    /// Runtime statistics accumulated across all LRA solves.
+    /// Runtime statistics for the current [`Self::solve`] call. Reset at the start of every
+    /// `solve` and accumulated across the LRA solves that call performs, so a reused
+    /// `LIRASolver` never carries stats — or the [`SolverConfig::max_lra_solve_calls`] budget
+    /// derived from them — over from an earlier `solve`.
     stats: Stats,
 }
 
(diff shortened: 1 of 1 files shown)


src/arithmetic/lia/lira_solver.rs
amarshah1: Is this still over all LRA solves or is it per call to the incremental solver?

src/arithmetic/lia/lira_solver.rs
```diff
@@ -133,6 +133,10 @@ impl LIRASolver {
     /// `INFEASIBLE` with a conflict core, or `UNKNOWN` if a resource limit was hit.
     pub fn solve(&mut self) -> SolverResult<SolverReturn> {
```
amarshah1, line 142: Could this be incorrect if we do not clear the internal state after an unsat result? If so, can we add some documentation to this.

amarshah1 (approved):
Looks good!

src/arithmetic/lia/lira_solver.rs
benjaminfjones: Good catch, I'll update the comment. It's per LIRA solver call now.

src/arithmetic/lia/lira_solver.rs
benjaminfjones: Another good catch, I think it could be incorrect. On reuse, solve() starts with try_unit_cube_test, which relies on set_backtrack()/backtrack() to scope its bounds. Both are no-ops while the underlying LRASolver is Unsat, so the cube test's tightened bounds would leak into the new solve. The old contract required the caller to call clear_unsat_state() (as the test did manually).

PR #192, opened 2026-08-17
Support div/mod terms in internal incremental arithmetic frontend

benjaminfjones:
*Description of changes:*

Followup to supporting incremental arithmetic in the internal arithmetic solver. This adds support for div/mod terms and adds many tests including tests of the assert -> solve -> backtrack -> solve -> etc. paths for both div/mod constraints and ordinary definitional equality constraints.

----


src/arithmetic/lia/incremental.rs
```diff
@@ -130,6 +133,9 @@ pub struct IncrementalLraSolver {
     next_var_id: VarId,
     /// `VarId` → internal solver [`Var`].
     var_of: DeterministicHashMap<VarId, Var>,
+    /// `VarId`s registered with a constant definition → that constant. Used to resolve
+    /// `div`/`mod` divisors.
+    const_of: DeterministicHashMap<VarId, IBig>,
```
amarshah1: It seems like this never gets cleared. It doesn't matter a ton now if we're not using the incremental arithmetic trait, but if we want to match the behavior of the Z3 incremental arithmetic, we would want that this gets cleared on a backtrack

amarshah1: But also not sure if that is the best way to do things, so we can talk about it more at the Sundance meeting or something

src/arithmetic/lia/incremental.rs
```diff
@@ -204,6 +233,9 @@ impl IncrementalLraSolver {
         }
 
         if let Some(def) = definition {
+            if let Some(c) = def.as_constant() {
+                self.const_of.insert(var_id, c.clone());
+            }
```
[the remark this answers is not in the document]
benjaminfjones: Good catch! Fixed now:

  - const_of now stores (IBig, usize), the constant plus the sat_level at which its defining equality was asserted
  - register_var records self.sat_level alongside the constant
  - notify_backtrack now drops every constant definition whose defining row was just relaxed. Those variables then correctly fail resolve_divisor (rejected as non-constant divisors)

src/arithmetic/lia/incremental.rs
```diff
@@ -328,30 +369,77 @@ impl IncrementalLraSolver {
...
+            // mod(a, n) = a - n*q
+            let c = Self::signed(coeff, negate);
+            monomials.push(Mon::new(c.clone(), a_var));
+            monomials.push(Mon::new(-(&c * &Rational::from(n)), q_var));
         }
         Ok(monomials)
     }
 
+    /// Push the Euclidean rows defining quotient `q` for `a / n` (constant `n`):
+    /// `a − n·q ≥ 0` and `a − n·q ≤ |n| − 1`, both tracked under `lit`.
```
amarshah1, line 415: what if lit = None, would this be a problem?

amarshah1 (approved):

src/arithmetic/lia/incremental.rs
benjaminfjones: This was fixed in the last commit I pushed (in response to Copilot)

src/arithmetic/lia/incremental.rs
benjaminfjones: If `lit == None` the definition pushed is treated as permanent, independent of what decisions the SAT solver is making. It also means it won't appear in unsat cores, which I think is consistent.

I think that's the right behavior, but correct me if I'm wrong. The impl goes through `push_relation` which has the same behavior for non-div/mod relations.

Issue #202, opened 2026-09-03
Non-interpretation of theory operators is unsound in general


benjaminfjones:
For example, currently Sundance treats `BitVector` operations as uninterpreted, leading to SAT results on UNSAT queries where we should return an error (unsupported operation).

The following is a small example where Sundance returns SAT despite the query being UNSAT and despite Sundance not supporting bit vectors:

```scheme
; Verus VC benchmark (solver flavor: cvc5)
; source log : operandabi_syscall_decode!operand.lemma_linux_int_from_abi_word_high_bits_ignored._01.smt2
; module     : operandabi_syscall_decode_operand.lemma_linux_int_from_abi_word_high_bits_ignored._01
; vc index   : 1
; expected   : unsat (Verus proved this VC)
;----------------------------------------------------------------------
(declare-sort %%Function%% 0)
(declare-const prefix! (_ BitVec 64))
(declare-const low_word! (_ BitVec 64))
(assert
 true
)
(declare-const %%location_label%%0 Bool)
(assert
 (not (=>
   %%location_label%%0
   (= ((_ extract 31 0) (bvor (bvshl prefix! ((_ zero_extend 58) (_ bv32 6))) (bvand low_word!
       ((_ zero_extend 32) (_ bv4294967295 32))
     ))
    ) ((_ extract 31 0) low_word!)
))))
(check-sat)

```

The same goes for `String`s, `Array`s, `Set`s, ...

benjaminfjones:
Addressed by #203

PR #203, opened 2026-09-03
Reject unsupported sorts and operators before they are registered in the egraph

benjaminfjones:
*Issue #, if available:* https://github.com/yaspar-org/Sundance-SMT/issues/202

----

### DIFF
--- a/src/theory_check.rs
+++ b/src/theory_check.rs
@@ -5,23 +5,40 @@
 //!
 //! Any operator that `SolverState::extract_op` does not recognise is registered in the egraph as an
 //! ordinary application (`Op::App`), so it is only constrained by congruence closure. For a theory
-//! operator that is unsound. Rather than answer incorrectly, we reject the input with an error
+//! operator, that is unsound. Rather than answer incorrectly, we reject the input with an error
 //! naming the theory and the offending term.
 
 use yaspar_ir::ast::{Context, FetchSort, Repr, Term};
+use yaspar_ir::statics::{BITVEC, STRING, ARRAY, REGLAN, SET};
 
-/// Sorts whose theory Sundance has no decision procedure for, paired with the
-/// human-readable theory name used in the diagnostic.
-const UNSUPPORTED_SORTS: &[(&str, &str)] = &[(yaspar_ir::statics::BITVEC, "fixed-size bitvectors")];
+/// The theory name for a sort Sundance has no decision procedure for, or `None` if the sort is
+/// supported.
+fn unsupported_theory(sort_name: &str) -> Option<&'static str> {
+    match sort_name {
+        BITVEC => Some("fixed-size bitvectors"),
+        STRING => Some("strings"),
+        REGLAN => Some("regular expressions"),
+        ARRAY => Some("arrays"),
+        SET => Some("sets"),
+        _ => None,
+    }
+}
 
 /// Reject `assertions` if any subterm belongs to an unsupported theory.
 ///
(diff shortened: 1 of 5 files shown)


src/theory_check.rs
```diff
@@ -0,0 +1,135 @@
...
+use yaspar_ir::ast::{Context, FetchSort, Repr, Term};
+
+/// Sorts whose theory Sundance has no decision procedure for, paired with the
+/// human-readable theory name used in the diagnostic.
+const UNSUPPORTED_SORTS: &[(&str, &str)] = &[(yaspar_ir::statics::BITVEC, "fixed-size bitvectors")];
+
+/// Reject `assertions` if any subterm belongs to an unsupported theory.
+///
+/// Detection is by sort rather than by operator.
+pub fn reject_unsupported_theories(
```
HuStmpHrrr, line 35: Cache to dodge checked terms?

HuStmpHrrr: Also what's the timing for calling it? Ideally it's the most efficient to call it even before letelim

src/theory_check.rs
```diff
@@ -0,0 +1,135 @@
...
+//! Any operator that `SolverState::extract_op` does not recognise is registered in the egraph as an
+//! ordinary application (`Op::App`), so it is only constrained by congruence closure. For a theory
+//! operator that is unsound. Rather than answer incorrectly, we reject the input with an error
+//! naming the theory and the offending term.
+
+use yaspar_ir::ast::{Context, FetchSort, Repr, Term};
+
+/// Sorts whose theory Sundance has no decision procedure for, paired with the
+/// human-readable theory name used in the diagnostic.
+const UNSUPPORTED_SORTS: &[(&str, &str)] = &[(yaspar_ir::statics::BITVEC, "fixed-size bitvectors")];
```
HuStmpHrrr: Consider using a phf map to avoid sequential lookup

src/theory_check.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Reject `assertions` if any subterm belongs to an unsupported theory.
+///
+/// Detection is by sort rather than by operator.
+pub fn reject_unsupported_theories(
+    assertions: &[Term],
+    context: &mut Context,
+) -> Result<(), String> {
+    let mut stack: Vec<Term> = assertions.to_vec();
```
HuStmpHrrr: I have a feeling that this stack can just be `Vec<&Term>` as you never actually need to own a term

src/theory_check.rs
benjaminfjones: already there: seen is a FastDeterministicHashSet keyed on `term.uid()`, so a shared subterm is checked once. It wasn't documented, so I added a comment

src/theory_check.rs
benjaminfjones: I added a small `unsupported_theory(sort_name) -> Option<&'static str>` helper. That removes the sequential scan without adding a phf dependency for what is currently a couple of static strings.

src/theory_check.rs
benjaminfjones: Yep, good point. fixed in the second commit

src/theory_check.rs
benjaminfjones: re: the timing, calls it immediately after typechecking, before the loop that does let_elim/NNF/CNF and before `SolverState::new`. I documented this on the function

HuStmpHrrr (approved):

Issue #183, opened 2026-08-10
Support Real Arithmetic


amarshah1:
Currently, we only support integer arithmetic. Real arithmetic support is doable, howere registering Real terms as-is would be unsound. The entire arithmetic pipeline assumes integers:

- All three backends allocate integer variables regardless of sort (`VarType::Int` in the internal solver, `Int::new_const` in the z3 backends).
- Nelson-Oppen buckets model values with an `IBig` (integer) key; Real models can be fractional, so distinct rational values would collide and merge terms that aren't equal.

PR #204, opened 2026-09-04
Reject non-linear arithmetic early

benjaminfjones:
*Issue #, if available:* related to #202

*Description of changes:*

This allows sundance to reject non-linear arithmetic before it hits the arithmetic solver layer, and exit cleanly. Currently presense of non-linear arithmetic causes a panic that happens during arithmetic theory solver translation that, because of Cadical integration, passes through the C++ FFI and becomes a SIGABRT.

----


src/main.rs
```diff
@@ -78,6 +78,12 @@ fn main() -> Result<(), String> {
     // problem.
     reject_unsupported_theories(&assertions, &mut context)?;
 
+    // Likewise for non-linear multiplication, which the arithmetic frontend cannot translate.
+    // Catching it here keeps it a clean error reported by the solver. The equivalent check inside
+    // `extract_linear_expression` runs in a CaDiCaL callback, where cxx turns the panic into
+    // `abort()`.
+    reject_nonlinear_arithmetic(&assertions)?;
```
amarshah1, line 85: Can you double check that we don't performance regress on anything?

src/theory_check.rs
```diff
@@ -60,14 +67,84 @@ pub fn reject_unsupported_theories(
...
+///
+/// * only `Constant::Numeral` counts -- a `Decimal` makes that function panic rather than fold
+/// * only `+`, `-` and `*` fold. `div` and `mod` become `Coefficient::Div`/`Mod` keys and `abs`
+///   and `/` fall through to the uninterpreted arm, so all four behave like variables even when
+///   applied to numerals
+/// * everything else (variables, `ite`, uninterpreted applications, annotated terms) becomes its
+///   own variable
+fn folds_to_constant(term: &Term) -> bool {
+    match term.repr() {
+        ATerm::Constant(c, _) => matches!(c, Constant::Numeral(_)),
```
amarshah1, line 84: I see so if we had Constant::Decimal, this would panice? I guess because we are not trying to support Reals right now, so its ok

src/theory_check.rs
```diff
@@ -60,14 +67,84 @@ pub fn reject_unsupported_theories(
...
+        _ => false,
+    }
+}
+
+/// Does `term` mention a variable bound by an enclosing quantifier, `let`, or `match` arm?
+///
+/// Such a term is not what the arithmetic frontend will eventually see: quantifier instantiation and
+/// let-elimination substitute for the bound variable first, and the result is often linear even when
+/// the body is not. See [`reject_nonlinear_arithmetic`] for why that forces us to defer.
+fn contains_bound_variable(term: &Term) -> bool {
```
amarshah1, line 98: maybe this should take in a cache, so that reject_nonlinear_arithmetic can be a linear algorithm and no quadratic

src/theory_check.rs
```diff
@@ -60,14 +67,84 @@ pub fn reject_unsupported_theories(
...
+///
+/// We reject any `*` with two or more factors that do not fold to a constant. That is precisely the
+/// condition under which `extract_linear_expression` panics.
+///
+/// **A product mentioning a bound variable is exempt**, even when it looks non-linear. The Verus
+/// prelude asserts `(forall ((x Int) (y Int)) (! (= (Mul x y) (* x y)) :pattern ((Mul x y))))`, and
+/// `(* x y)` is never handed to the arithmetic frontend as written, only its instantiations
+/// are. Rejecting on the body would reject essentially every Verus input.
+///
+/// The cost of that exemption is that this check is *not* complete: an instantiation can still
```
amarshah1, line 112: Couldn't this also happen because of inlining lets/defin-fun. Since `reject_nonlinear_arithmetic` comes before the inling step

amarshah1 (approved):
Looks good! If you address the comments, you can merge this PR in

src/main.rs
benjaminfjones: I measured on a random selection of the corpus, no regression. Release builds of main (876a8b8) vs the branch over 400 random Verus files (10s timeout): got identical top-level results on all 400 (309 unsat, 3 unknown, 88 timeout, 0 diffs); total wall time 974.7s vs 973.5s, and 93.6s vs 92.5s (−1.2%, noise) over the 312 files that finished.

src/theory_check.rs
benjaminfjones: Yeah, it would, at lp.rs:378 ("non-numeric constant in arithmetic expression"), but in practice nothing gets that far. A Real-sorted term is stopped earlier in solver_state.rs:531 with "Real arithmetic is not yet supported". We'll have to update both when real arithmetic is supported. cf. https://github.com/yaspar-org/Sundance-SMT/issues/183

src/theory_check.rs
benjaminfjones: Good idea! Fixed this before running the performance comparison commented above.

src/theory_check.rs
benjaminfjones: Yep, that's a gap. Thanks. 

`(let ((y x)) (* x y))` and a `define-fun` body were both exempt (bound variable / not in any assertion), then became ground non-linear products after inlining and aborted at solve time. The check now runs inside the preprocessing loop in main.rs on `assert.let_elim(..).gsubst_all(..)`, so it sees exactly what the arithmetic frontend will.

I also added some additional unit tests.

Issue #134, opened 2026-06-24
Support integer mod in the arithmetic solver


benjaminfjones:
In order for integer division/modulo to be supported in Sundance, the arithmetic solver needs to handle `mod` terms like it already does `div` terms.

amarshah1:
#141 gave this fix by adding an axiom for integer mod. @benjaminfjones do you think this is sufficient or do we want to support mod as an operation inside the internal solver?

benjaminfjones:
I think the axiomatic approach is sufficient for now. In the meantime I'll look into how other implementations handle it.

PR #205, opened 2026-09-04
Support integer mod in the internal arithmetic frontend

benjaminfjones:
*Issue #, if available:* https://github.com/yaspar-org/Sundance-SMT/issues/134

*Description of changes:*

Support integer `mod` in the internal arithmetic frontend.

Encoding: for `(mod e n)` with `n` a nonzero constant, it allocates a remainder var `r` (associated with the mod term itself, so it shows up in models) plus an auxiliary integer quotient `q`, then pushes three relations matching SMT-LIB Euclidean semantics:

  - `e - n*q - r = 0`
  - `r >= 0`
  - `r <= |n| - 1`

  and returns r as the linear expression. Non-constant divisors error with "non-linear integer modulus is not supported"; a zero divisor errors, mirroring div.

The change also fixes a latent bug in the integer `div` encoding where constraints were being duplicated.

----


src/arithmetic/lia/frontend.rs
```diff
@@ -293,6 +299,80 @@ fn convert_linear_term(ctx: &mut ConvContext, term: &Term) -> FrontendResult<Lin
...
+                            return Err(FrontendError("integer modulus by zero".to_string()));
+                        }
+                        // The remainder variable is associated with the mod term itself, so
+                        // that it is reported in models; the quotient is auxiliary.
+                        let r = ctx.allocate_term(term.clone(), None, app_sort.clone());
+                        let q = ctx.allocate_var(
+                            &format!("$mod_quotient_{}", ctx.num_variables()),
+                            VarType::Int,
+                        );
+                        let abs_n = if n < Rational::ZERO {
```
amarshah1: can Rational just expose an `abs` function?

amarshah1 (approved):
Looks good!

src/arithmetic/lia/frontend.rs
benjaminfjones: yep, I forgot there was an instance

src/arithmetic/lia/frontend.rs
```diff
@@ -293,6 +299,80 @@ fn convert_linear_term(ctx: &mut ConvContext, term: &Term) -> FrontendResult<Lin
...
+                        if n.is_zero() {
+                            return Err(FrontendError("integer modulus by zero".to_string()));
+                        }
+                        // The remainder variable is associated with the mod term itself, so
+                        // that it is reported in models; the quotient is auxiliary.
+                        let r = ctx.allocate_term(term.clone(), None, app_sort.clone());
+                        let q = ctx.allocate_var(
+                            &format!("$mod_quotient_{}", ctx.num_variables()),
+                            VarType::Int,
+                        );
```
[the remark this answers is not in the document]
benjaminfjones: fixed

